PR #8, opened 2022-11-08
fft

nupur-lal:
Added demo on Discrete Fast Fourier Transform


DougEbel (changes requested):
The name of the notebook should be meaningful and match the directory name.  The user may be browsing the directory list provided on the left by Jupyter and FourierTransform.ipynb will be more logical than fft.ipynb.

PR #49, opened 2022-12-15
Val analytic func

shilpa-nalkande:
@DougEbel . Updated the Val Notebooks. Please review.


DougEbel:
When I switch to branch, the %connect fails on the SQL notebooks. Says there is no local connection.  Details sent in e-mail.

PR #50, opened 2022-12-15
nyc standardized notebooks

nupur-lal:


DougEbel:
Recommend combining Timeseries into 4d and naming both ... Timeseries.   The 4d has no x/y/z coordinates, only the time demension.

PR #88, opened 2023-01-18
Val pass3

shilpa-nalkande:
@DougEbel . Changes made for Pass 3 in VAL notebooks


DougEbel:
Rejected ... other notebooks in pull request than just VAL

PR #117, opened 2023-01-30
Parkinsons pass3

shilpa-nalkande:
@DougEbel . Created another pull request with only the Parkinsons file


DougEbel:
Thanks for narrowing down to 1 file.  However, this has inconsistent story, poor performance and still hoping we can improve the prediction.

PR #133, opened 2023-02-09
pass3 tf_idf for pull req

nupur-lal:


DougEbel:
The visualization at the end for SQL isn't sorted. In Python it is.

PR #155, opened 2023-03-13
Sales forecasting

shilpa-nalkande:
Sales Forecasting Demo using ARIMA


DougEbel:
Comments sent by e-mail on 3/15 ... eliminate "walmart", sorting of output, precision of numbers, and explain why estimate high.

PR #156, opened 2023-03-14
Banking Churn using BYOM

ForbiddenDevil:
@DougEbel Please review the new usecase of Banking Churn.


DougEbel:
Failed - Email sent regarding: missing: from xgboost import XGBClassifier

DougEbel (commented):
The performance was great!!  
•	However, there was still a pip install of xgboost which happened before the session was recorded in qrylog which was a bit slow.
•	There aren’t any visualizations until the ROC curve at the very last step.

PR #166, opened 2023-03-23
DBQL new Notebook

shilpa-nalkande:
Created DBQL notebook with points mentioned in mail


DougEbel (commented):
The term "Default Row" is strange.  It implies that the row can be changed or optioned to be something else. It is merely the defined row.  

The comment to double click the output of SQL text was copied from the cell below which was for the explain text.

DougEbel (approved):
I don't get any output for 8 because somehow earlier session steps weren't there.   But it worked for other sessions.

PR #191, opened 2023-04-10
New notebook for Green Manufacturing

shilpa-nalkande:
Green Manufacturing notebook

### DIFF
--- a/UseCases/GreenManufacturing/GreenManufacturing.ipynb
+++ b/UseCases/GreenManufacturing/GreenManufacturing.ipynb
@@ -1,0 +1,1189 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "ccb0e42b-1d06-4ae0-91ec-e06fd55d6d3d",
+   "metadata": {},
+   "source": [
+    "<header style=\"padding:1px;background:#f9f9f9;border-top:3px solid #00b2b1\"><img id=\"Teradata-logo\" src=\"https://www.teradata.com/Teradata/Images/Rebrand/Teradata_logo-two_color.png\" alt=\"Teradata\" width=\"220\" align=\"right\" />\n",
+    "\n",
+    "<b style = 'font-size:28px;font-family:Arial;color:#E37C4D'>Green Manufacturing for vehicles</b>\n",
+    "</header>\n",
+    "\n",
+    "<p style = 'font-size:20px;font-family:Arial;color:#E37C4D'><b>Introduction</b></p>\n",
+    "\n",
+    "<p style = 'font-size:16px;font-family:Arial'>Since the first automobile, the Real Wheel Motor Company has stood for important automotive innovations. These include, for example, the passenger safety cell with crumple zone, the airbag and intelligent assistance systems. Real Wheel Motor Company applies for nearly 2000 patents per year, making the brand the European leader among premium car makers. These cars are leaders in the premium car industry. With a huge selection of features and options, customers can choose the customized models of their dreams.</p>\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


mrconway (changes requested):
Please address any issues as needed, taking into account the resolution of the reviewer's commentary. Some of these items have already been addressed by the answers in _(italics)_.

1. Start by connecting to the Vantage system.
Just verifying that the Python interpreter restart issue can be resolved by the startup code. _(We have already discussed these points and concluded that we need a RESTART)_

2. Getting Data for This Demo
Move the anticipated timing from Python comments to bold Markdown. _(Standards followed in all notebooks, where we specify different time for CLOUD and LOCAL)_

General Comment
Just as a general observation, there’s some concern regarding the notebook startup phase in terms of package installation and data loading, in terms of detracting from the user experience. This part should be as transparent as possible to get to the fun stuff. _(For data loading we have created Stored Procedure to standardize the process and minimize activities in the notebook)_

3. Analyze the raw data set
The features and target variables should be in tabular format. Is there any equivalent to the pandas-profiling package here, where we can quickly look at distributions, missing values, etcetera? _(Not required for the flow of the use case and keep the demo concise)_

4. Check the impact of Categorical variables on target variable 'y’
We may want to put in an ANOVA analysis here to appeal to statisticians, that is, If the p-value is below a chosen significance level (e.g., 0.05), then there is evidence to suggest that the means of the target variable are significantly different across categories. _(This is only EDA- initial analysis of the data)_

5. Check the importance of various features on target variable 'y’
While the feature importances of XGBoost are sufficient, we should also display Shapley visualizations such as waterfall, swarm, and force plots. Models need to be interpreted and explained in a business context; otherwise, feature importances in isolation are not as informative. _(I am not sure if Shapley works with Teradata models, and also we do not have any function or method to check feature importance in Vantage.)_

6. OrdinalEncoding of the categorical variables
We should explain the difference between normal categorical variables and ordinal variables because of the implicit ordering, i.e., what are the implications for not accounting for order, and how do machine learning algorithms handle these types of situations? _(We can add the explanation for Categorical and Ordinal variables. Will look into this and send the explanation before actually making changes in notebook)_

7. Preparation of Data
The important features are manually entered in a Python list, when they should actually be extracted programmatically from the XGBoost model, e.g., top N features (either as a number or percentage). _(Cannot be done automatically as feature importance is not available in Vantage)_

8. Decision Forest:- _(The description was added from TD documentation, but can make below changes )_
Here’s an alternative description for this section:

“The Decision Forest is a powerful method used for predicting outcomes in both classification and regression problems. It's an improvement on the technique of combining (or "bagging") multiple decision trees. Normally, building a decision tree involves assessing the importance of each feature in the data to determine how to divide the information.
This method takes a unique approach by only considering a random subset of features at each division point in the tree. This forces each decision tree within the "forest" to be different from one another, which ultimately improves the accuracy of the predictions. The function relies on a training dataset to develop a prediction model. Then, the TD_DecisionForestPredict function uses the model built by the TD_DecisionForest function to make predictions. It supports regression, binary, and multi-class classification tasks.

Please note the following points:
1.	All input features should be numerical. If you have categorical columns, convert them to numerical columns as a preprocessing step.
2.	For classification, class labels (the values in the ResponseColumn) must be integers. There's a limit of 500 classes for classification tasks.
3.	Observations with missing values in any input column will be ignored during training. To fill in missing values, use the TD_SimpleImpute function.
4.	The number of trees built by the TD_DecisionForest function depends on the values of NumTrees, TreeSize, and CoverageFactor, as well as the data distribution in the cluster. The trees are built simultaneously by all the processing units (AMPs) that have a non-empty portion of the data.”

9. XGBoost
Let’s evaluate the metrics in context of the data we’ve explored
•	What did we learn about our predictive model?
•	How do the results translate to the real world?
_(In the end I have concluded, which model is working better , not sure how we can translate to real world, Would be helpful if I can get some more details on what is expected here.)_

PR #192, opened 2023-04-10
carbonfootprintanalytics added review comments

nupur-lal:


DougEbel (commented):
This is better than what was out there.  The last graph is confusing.  It is hard to believe that we did almost no carbon in 2021.  Also the Y axis numbering makes no sense.   I'll merge it but leave it out there for improvement of the last graph.

DougEbel (approved):
This is better than what was out there.  The last graph is confusing.  It is hard to believe that we did almost no carbon in 2021.  Also the Y axis numbering makes no sense.   I'll merge it but leave it out there for improvement of the last graph.    

The image that I requested to be replaced was supposed to be replaced with a mark-down equivalent.  Not only was it blurry, but Chris didn't like the use of the Microsoft Icons on the right.      

This is merged and the branch is closed.  Create a new branch for added changes (?carbon_11apr?)

PR #193, opened 2023-04-11
Added new usecase Predictive Maintenance

ForbiddenDevil:
@DougEbel  Please review the new usecase Predictive Maintenance. Thanks.


mrconway (changes requested):
This review contains the original viewer comments in the first section **and** the review of the reviewer comments in the latter section. Request changes that can be reasonably accommodated.

**Reviewer Comments**

1.	Introduction: “operates a large fleet of machines” 
>>> What type of machine? This will pique interest in the subject.

2.	Introduction: “deploy machine learning algorithms through teradataml”
>>> Mention that it’s a Python library.

3.	Introduction: Imports
>>> I like that the imports are alphabetized, but we should probably separate the variable assignments that follow into another section.

4.	Getting Data for This Demo
>>> Instead of asking the user to uncomment one line and comment out the other, use a flag (True/False) to select the proper option:

use_cloud_storage = True
if use_cloud_storage:
    # do this
else:
    # use local storage

5.	Section 2: “The dataset above has ten columns, and the 'Target' and 'Failure_Type' columns are dependent columns where the Target column has 1 (Failure) and 0 (No Failure). The Failure_Type column has multiple types of failures.”
>>> I would distinguish here between a Binary Classification model (1/0) and a MultiClass Classification model.

6.	Section 3: Data Exploration: “In the next cell, we'll remove the null values from the dataset.”
>>> What happens if most of the data is there, but a single column has mostly null values. This means most of the rows, even with meaningful information, will be dropped. I’m not sure we have to discuss imputation here, but we should be displaying counts of null values.

7.	Section 3: Data Exploration: Checking Target Variable Distribution
>>> Good walkthrough here

8.	Section 3: Data Exploration: Checking the Correlation
>>> My favorite use of Seaborn using hue for contrast. We can support these insights later with a Feature Importance analysis, and this will confirm the insights stated here. Note that we can bolster many of our ML notebooks with packages for Shapley Values such as shap, as the visualizations are quite compelling.

9.	Section 4: Data Transformation
>>> Add an explanation for Z-Score during the label encoding discussion.

10.	Section 4: Data Transformation
>>> Delineate a separate section for the train/test split.

11.	Section 5: In Database Model Training (Binary Classification)

“The TD_XGBoost function, eXtreme Gradient Boosting, implements the gradient-boosted decision tree designed for speed and performance. It has recently been dominating applied machine learning.
…
of the previous iteration to correct the errors made by existing models. The predicted residual is multiplied by this learning rate and then added to the previous prediction. Models are added sequentially until no further improvements can be made. It is called gradient boosting because it uses a gradient descent algorithm to minimize the loss when adding new models.”

>>> Add an explanation why XGBoost works better on tabular data. Plus, I think there was some deleted text here in the middle.

12.	Section 6: In Database Model Scoring
>>> Let’s add an analysis of the Confusion Metric, plus we can supplement it with a combination of calibration curve, learning curve, ROC/AUC, or feature importance plots.

13.	Section 8: In Database Model Scoring (for Multiclass Classification)
>>> See previous comment – there are plenty of good classification visualizations to compare and contrast.

14.	Conclusion
>>> What’s missing here is a translation (i.e, a mapping of the model results) to the business bottom line. For example, how long can they expect a certain part to last, and what is the replacement policy. Any machine learning model directly translates to the business bottom line, for example, increase in revenue or cost savings. Ultimately, we need to convince the customer that a model is going to help them.

**Review Response**

1.	I am not aware what kind of machine’s data this is. Pranay provided the data from Kaggle. I checked on multiple sources, I didn’t find what machine’s data is this.
2.	I’m happy to get such markup suggestions as this gives a chance to improve from a 3rd person Point Of View.
3.	This is something that if we change, we might need to change in all the notebooks for consistency purpose. I don’t think this makes a huge difference in making the notebook better.
4.	This is surely a better way to do it, if you agree with this. We need to change this is all the notebooks. Shouldn’t take much time.
5.	This suggestion is welcomed.
6.	The data didn’t contain any null values hence I did not show null values neither did any imputation. If required in any usecase, I will be using these techniques.
7.	Thank you.
8.	Thanks for the complement on visualization. I haven’t used “Shapely shap” for model explanations. I’ll definitely learn this to use it and make our notebooks better. My only concern is: I’m afraid this “Shapely shap” is not compatible with terdataml models. I’ll comment more on this once I understand the library.
9.	Suggestion is welcomed.
10.	Minor change. Doesn’t improve the notebook.
11.	Suggestion is welcomed. I guess I missed some text here. Thank you.
12.	Confusion matrix is already added. More analysis can be added.
13.	Confusion matrix is already added. ROC/AUC is supported only for binary classification. More analysis can be added.
14.	This demo gives data just enough to say:
a.	whether the part will fail or not
b.	What type of failure it is

ForbiddenDevil:
@mrconway Please review the updated notebook. I was able to incorporate point number 2, 3, 5, 6, 7, 9, 10, 11, 12, 13. Thank you for the guidance.

mrconway:
> @mrconway Please review the updated notebook. I was able to incorporate point number 2, 3, 5, 6, 7, 9, 10, 11, 12, 13. Thank you for the guidance.

Great, will do.

mrconway (approved):
First of all, great job with all the new visualizations. The new scatterplots are really illustrative.

Only one comment below:

**9. In Database Model Scoring**

Let's explain how the model predicts for some aspects of the confusion matrix. For example, when the true value is Heat Dissipation, how well do we do predicting on the failures? In this case, we have 1 instance where Heat Dissipation was misclassified as Power Failure.

Approved, but let's add this one explanation above as well.

PR #196, opened 2023-04-12
Telco churn

nupur-lal:


mrconway (approved):
Review Summary:

- All cells ran successfully and in the expected amount of time.
- Very thorough Exploratory Data Analysis with all plots displaying as expected
- Model interpretation with ROC Curves for logistic regression and gradient boosting is good basis for comparison

We could use an explanation of Mean Absolute Error as it relates to the model results because it's just a metric in isolation that doesn't relate to anything else around it. There may be a little more refinement needed, but we are good to go here.

PR #205, opened 2023-04-19
Updated FSCustomerJourney to python and rename

shilpa-nalkande:
Updated the changes recommended by Pranay


mrconway (changes requested):
Please refer to the following comments:

- **Section 3: Data Exploration** explain the column names of the _Custdf_ dataframe.

- **Section 4: Customer Acquisition** briefly describe how Uniform Attribution works

- **Section 4: Customer Acquisition** correct the y-label "_Marketing Description_" of the Average Attribution chart

- **Section 4: Customer Acquisition** mention that the average attributions are percentages that sum up to 100%

- **Section 4: Customer Acquisition** For the chart "Total Attribution by Type of Promotion", what do the numbers actually mean, as the totals sum to greater than 100%. Exactly what is an attribution score?

- **Section 5: Channel Analysis** For the chart "Channel Performance", what does the black horizontal line through Google mean?

- **Section 5: Channel Analysis** For the next "Channel Performance" chart with the circles, I can't really match the circles on the right to the marketing channels on the left. It's unclear that this is the right visualization because the circles will naturally grow larger as you scan to the right of the chart.

- **Section 6: Path to Adoption** I would add a couple of sentences about the model or functionality behind the nPath analytic function.

- **Section 6: Path to Adoption** The query _query3_ took about 25 seconds when running this cell, maybe put a note about this.

- **Section 6: Path to Adoption** The function _plot_first_main_paths_ took ~48 seconds to run.

- **Section 6: Path to Adoption** I can't follow the Sankey flows. We should probably decompose into several flows.

- **Section 6: Path to Adoption** What's interesting about the Sankey flows is that the path widths are pretty uniform to the end. What makes Sankey so informative is the varying widths of different paths (like the infamous [Napoleon's Russian Campaign map](https://ageofrevolution.org/200-object/flow-map-of-napoleons-invasion-of-russia/)), and we don't see that here.

- **Section 8: Dataset** Okay, we have descriptions of the dataset fields here. We could insert a hyperlink from Section 3 referencing them.

PR #207, opened 2023-04-20
Renamed the usecase and enhanced the notebook

ForbiddenDevil:
@DougEbel , @mrconway Please review the updated notebook. I have renamed the usecase and folder from "UseCases/GLM_Fraud_Detection_InDB/GLM_Fraud_Detection_InDB.ipynb" to "UseCases/Financial_Fraud_Detection_InDB/Financial_Fraud_Detection_InDB_SQLPY.ipynb".

Specifications document: [here](https://teradata.sharepoint.com/:w:/r/teams/PortfolioMarketingTeam/Shared%20Documents/_PRODUCT%20MARKETING/Pranay%27s%20working%20docs/ClearScape_Analytics_Experience_Platform/Functional_specs/FINANCE_FINANCIAL%20TRANSACTION%20FRAUD%20DETECTION%E2%80%8B.docx?d=w6cae0ee94d5144b6b7333a6134d64203&csf=1&web=1&e=UHxxB0)

Kanban card is not created for this one.


mrconway (changes requested):
Please find comments below:

**3.1 Renaming columns**

How does renaming a column relate to transferring funds from one account to another?

**3.2 How many fraudulent transactions do we have in our dataset?**

Is a chart really useful here, or should we simply point out that 14 out of 10,000 transactions are fraudulent? Maybe both.

**3.4 What percentage of fraudulent transactions do we have where transaction amount is equal to old balance in the origin account?**

Please refer to comment for Section 3.2.

**4.1 Drop unneccessary columns**

Correct to "unnecessary"

**4.2 One-hot encoding**

Just a brief explanation of why one-hot encoding is necessary here, for example, generally it's used to convert all categorical columns to numeric representations.

**4.3 Feature Scaling**

_"Scaling of data makes it easy for the model to learn and understand the problem."_

In a sense, yes, but the main two reasons are:

1. Scaling features restrict models from being biased towards features having higher/lower magnitude values, and
2. Scaling helps optimization algorithms reach local optima, especially in gradient descent.

_5.1 Generalized Linear Model_

Great explanation, I enjoyed reading this section. Learning something new is always good.

For the feature importances, do we know why the flag _isFlaggedFraud_ has zero importance? Ah, wait a minute, is this the target variable for prediction? If so, then this needs to be removed from the training data. Please clarify.

We have a table of model output, but all of the items are lumped together. Consider this summary to describe our table:

1. Model Configuration:

- Loss Function (LOG): The logarithmic loss function measures the difference between the predicted probabilities and the actual outcomes. It is commonly used in binary classification problems.

- Regularization (Enabled, 0.02): Regularization is a technique used to prevent overfitting by adding a penalty term to the loss function. The value 0.02 indicates the strength of this penalty.

- Alpha (Elasticnet, 0.15): Alpha is a parameter in the Elastic Net regularization method, which combines L1 (Lasso) and L2 (Ridge) regularization. A value of 0 represents pure L2 regularization, while a value of 1 represents pure L1 regularization. In this case, the model uses a combination of both (0.15).
  
2. Optimization Algorithm:

- Learning Rate (Initial, 0.05; Final, 0.388190): The learning rate determines the step size the optimization algorithm takes during each iteration. The initial learning rate is set to 0.05, and it reaches a final value of 0.388190.

- Momentum (0.0): Momentum is a technique used in optimization algorithms to speed up convergence by adding a fraction of the previous update to the current update. A momentum value of 0.0 means it's not being used in this case.

- Nesterov (FALSE): Nesterov momentum is a variation of the standard momentum method. This parameter being set to FALSE indicates it is not being used.

3. Model Performance:

- Log-likelihood (-0.013004): Log-likelihood is a measure of how well the model fits the observed data. Higher values indicate a better fit.

- Akaike Information Criterion (AIC, 26.026008): AIC is used to compare models with different numbers of parameters, balancing goodness of fit with model complexity. Lower values indicate a better model.

- Bayesian Information Criterion (BIC, 140.9153): BIC is similar to AIC but places a stronger penalty on model complexity. Lower values indicate a better model.

4. Convergence:

- Number of Observations (50,901): The total number of data points used in the model.

- Number of Iterations (111, Converged): The number of iterations the optimization algorithm took to converge, indicating the algorithm has found an optimal solution.

- Intercept (-4.113711): The intercept (or bias term) is the value of the output when all input features are set to zero.

- LocalSGD Iterations (0): The number of iterations used in the Local Stochastic Gradient Descent (SGD) method. A value of 0 indicates it's not being used in this case. 

**5.2 XGBoost**

This query took about 20 seconds -- may want to mention.

**6.1 Generalized Linear Model**

This is tricky because we have an extremely imbalanced dataset where accuracy is not meaningful, it's locating the proverbial needle in the haystack, that is, how good are we finding at least some of the fraudulent transactions. I don't see that here because this is really the only finding that is relevant.

**6.2 XGBoost**

The same comments as in 6.1 apply, although I do see one case where the model identified a fraudulent transaction.

**7.1 Comparing the metrics**

Yes, the real highlights are the Macro-Precision, Macro-Recall, and Macro-F1 scores. Good explanation, well done.

ForbiddenDevil:
@mrconway According to your mail on 24-Apr, I have added the model output information at the end of the notebook. Please review it.

Commit: https://github.com/Teradata/jupyter-demos/pull/207/commits/0996c58292490b9f4624efe3e8ec7f8dd70c6fec

mrconway (approved):
ROC Curves illustrate beautifully. Great final product.

PR #216, opened 2023-04-25
Index 2023 04 23

DougEbel:
This is the updated index with renamed directories and notebooks named to correspond to the index and append with the language type. SQL notebooks with %connect missing the hidewarnings=true were fixed. The DDL directory for the legacy data dictionary was moved under UseCases/Data_Dictionary


mrconway (approved):
Merge new index.

PR #230, opened 2023-04-26
Hear Failure Survival Analysis

ForbiddenDevil:
@DougEbel Please rename this usecase and folder to **Heart Failure Analysis**. As per Pranay, Survival Analysis is a different technique altogether which we are not using.

@mrconway Please review the code. The TD_GLM function does not have seed parameter and hence the outputs vary slightly on each run. I'm open to suggestions. Thank you.


mrconway (changes requested):
Please find commentary below:

**Introduction:**

For the last bullet point, it looks like some text got cut.

"_By 2035, more than 45 percentTrusted Source of Americans are projected to have some form of cardiovascular disease_."

**2. Data Exploration**

Show a count of deaths vs. non-deaths.

**2.1 Mortality Rate by Gender**

We concluded that "_gender does not appear to be a determining factor in mortality rates_" here, but perhaps we can prove this statistically with a t-test or similar test. It's not required, but just an idea.

**2.3 Correlation Matrix**

Let's point out some negative correlations as well to illustrate the difference between the colors.

**2.4 Pairplot for multivariate correlations**

As you mention, those particular pairs are inconclusive, but we know this from the previous plot. To show correlation, let's take one or two of the correlated pairs from above (1 positive correlation and 1 negative correlation) to illustrate the difference.

**2.5 Distribution plots for numeric variables**

I think we need more explanation here. For example, the first distribution is exponential, and why that may be so.

**2. Data Prepration**

"_2. Data Prepration_" -> _Data Preparation_

Change from: _Here, we classify the columns in to ID column, numerical, categorical and binary ones._ to: "_classify the columns into an ID column,_ ..."

_Keep the other variables that do not not need trasformation._ -> "_transformation_"

**3. Model Training**

_The feature importances displayed above indicate that age and the number of appointments missed are significant factors in predicting heart failure deaths. Other notable factors include the presence of arrhythmia (irregular heartbeat), hypertension, COPD (chronic obstructive pulmonary disease), and whether the patient belongs to the Indian subcontinent._

Although you explain the impact of the various feature importances, we should note that the top ones negatively affect the target variable and that the bottom ones positively affect the target variable.

**4. Model Validation**

No changes needed in this section. This is really well done.

**5. Visualize the results**

Let's explain one aspect of the confusion matrix in words. For example:

When the outcome was Death, our model correctly predicted x% of the total, and when the outcome was Not Death, our model correctly predicted y% of the total.

ForbiddenDevil:
- Improved the introduction text to provide a better overview of the project.
- Added counts for both death and non-death cases to provide a better understanding of the data.
- Skipped the t-test as the graph was indicative enough.
- Added negative correlation examples to the description to highlight the relationships between variables.
- Determined that the pairplot was not particularly useful due to the binary nature of many of the correlated variables, but still included it as per the requirements.
- Added descriptions to the description plots for greater clarity.
- Corrected several typos in the data preparation stage to ensure accurate data processing.
- Added a note during model training regarding the positive and negative effects of variables according to their coefficients.
- Included an explanation of the confusion matrix to provide a better understanding of the model's performance.

mrconway (approved):
All changes made as requested. Approved.

PR #231, opened 2023-04-28
Removed redundant cell from cleanup section

ForbiddenDevil:
@mrconway @DougEbel Removed the cleanup cell from the cleanup section of Data Prep and Transformation usecase as per you mail on Fri 28-04-2023 09:13 IST.


mrconway (approved):
Ran full notebook, fix confirmed.

PR #233, opened 2023-04-28
Review and minor changes

ForbiddenDevil:
@DougEbel @mrconway Please review the changes in these notebooks. These notebooks run fine.


mrconway (approved):
Just as a precaution even for minor changes, we'll always run the notebooks, read them, and confirm the output.

Good to see the different ways of doing TF-IDF!

PR #246, opened 2023-05-09
 Python converted demo for Financial Fraud InDB

ForbiddenDevil:
@mrconway @DougEbel This is the python version provided by Product Team(Gauri). Please have a look and suggest any changes if required. Thanks.

### DIFF
--- a/UseCases/Financial_Fraud_Detection_InDB/GLM_Fraud_Detection_InDB-Python.ipynb
+++ b/UseCases/Financial_Fraud_Detection_InDB/GLM_Fraud_Detection_InDB-Python.ipynb
@@ -1,0 +1,647 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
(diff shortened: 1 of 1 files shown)


DougEbel:
naming wasn't consistent with folder and use - instead of _ before language.  I've renamed the file from "UseCases/Financial_Fraud_Detection_InDB/GLM_Fraud_Detection_InDB-Python.ipynb" to "UseCases/Financial_Fraud_Detection_InDB/Fraud_Detection_InDB_Python.ipynb" and will merge into main separate from this pull request. 

There was a !pip install --upgrade teradataml that I commented out. 

I added:
*RESTART the kernel if you uncomment the !pip statement above to bring new software into Jupyter.

The above statements may need to be uncommented if you run the notebooks on a platform other than ClearScape Analytics Experience that does not have the libraries installed. If you uncomment those installs, be sure to restart the kernel after executing those lines.

PR #248, opened 2023-05-09
 Python converted demo for Financial Fraud BYOM

ForbiddenDevil:
@mrconway @DougEbel This is the python version provided by Product Team(Gauri). Please have a look and suggest any changes if required. Thanks.


DougEbel:
At the end, and exception is raised on db_drop_table(table_name='clean_data_train')

Also, the name of the notebook is not consistent with the folder name or the otherr notebook name.

DougEbel:
This was not merged due to the exception.

DougEbel:
This also had 2 pip installs that were in the current VM image:

    "!pip install sklearn2pmml --user\n",
    "!pip install jdk4py --user\n",

PR #258, opened 2023-05-12
Anomoly detect

shilpa-nalkande:
Updated review comments and Renamed


DougEbel:
Review comments from May 9 were ignored and the change to eliminate the #TODO caused all steps after that in the notebook to fail.

mrconway (changes requested):
Hi Shilpa:

Please add the following explanatory text to the introduction:

"To apply anomaly detection techniques to spot welding, we:

1. **Data Collection**: Collect data related to our welding process. This includes the voltage, current, and resistance during the welding process, the time it takes for each weld, the thickness of the materials being welded, and any other relevant factors. This data is gathered from sensors attached to the welding equipment or from the equipment's control system.

2. **Data Preprocessing**: Clean and preprocess our data. This involves removing outliers or erroneous readings, handling missing values, and normalizing the data. Depending on the machine learning algorithm you plan to use, you may also need to transform the data in some way, such as rescaling it or converting it to a different format.

3. **Feature Extraction**: Determine what features or combinations of features are most indicative of a good or bad weld. We can use statistical analysis, machine learning techniques, or domain knowledge. For example, a sudden increase in resistance might indicate a poor weld.

4. **Model Training**: Here, we take our preprocessed data and train a machine learning model to recognize normal and abnormal patterns in our data. We can use clustering algorithms like K-means, classification algorithms like Support Vector Machines (SVM), or neural network-based approaches like autoencoders. The primary objective is to classify whether a weld is good or bad, with 0 representing no anomaly and 1 representing that an anomaly is present, i.e., a bad weld.

5. **Anomaly Detection**: Now, we use our trained model to identify anomalies in new welding data. This could involve running the model in real-time as the welding process is happening, or it could involve periodically analyzing batches of collected data. Either way, the goal is to identify when the data deviates from what the model considers normal, which could indicate a problem with the welding process.

6. **Evaluation and Adjustment**: Regularly evaluate the performance of our anomaly detection system and adjust it as necessary. We can retrain the model with new data, adjusting the sensitivity of the anomaly detection, or changing the features that the model considers."

You may also want to add explanation about the welding process:

The graph you mentioned, showing the variation of the resistance of the welding with respect to time, and the subsequent calculation of the derivative, is a common practice in analyzing and monitoring the welding process. 

This is done for several reasons:

1. **To Monitor Quality**: Resistance is an important parameter in resistance welding. The resistance should not vary too much. If there are any significant changes in resistance over time, it could indicate an issue with the weld quality. For example, an unusually high resistance could indicate poor contact between the parts being welded or a problem with the welding equipment.

2. **To Detect Anomalies**: By monitoring the resistance over time and calculating its derivative, you can detect any sudden changes or anomalies. Anomalies might indicate a problem with the welding process, such as a sudden drop in current or a sudden increase in resistance.

3. **To Predict Issues**: Monitoring the resistance and its changes over time can help to predict potential issues before they become serious problems. For example, if the resistance starts to gradually increase over time, it could indicate that the welding electrodes are wearing out and need to be replaced.

4. **To Understand the Welding Process Better**: By studying how resistance varies over time during the welding process, you can gain a deeper understanding of the welding process itself. This can help to optimize the process, improve weld quality, and increase efficiency.

Applying a window function to smooth the resistance curve helps to eliminate noise and makes it easier to see the overall trend. The derivative of the resistance gives an indication of how quickly the resistance is changing, which can be a useful measure for detecting anomalies and predicting potential issues.

DougEbel (approved):

PR #262, opened 2023-05-15
AirPassengers: Changes made according to updated Dickey_Fuller Test

ForbiddenDevil:
@DougEbel @mrconway The changes are made according to the updated Dickey-Fuller test results. More descriptive text has also been added. Please review. Thanks.


mrconway (approved):
Dickey-Fuller test produces the right output.

Approved

PR #264, opened 2023-05-16
uaf cash demand forecasting

nupur-lal:
cash demand forecasting notebook using UAF functions in sql kernel, dropping the other branch 'uaf' as it has some file of another usecase added in it


DougEbel:
You ought to describe the number of rows being processed.  At the top, I was lead to believe there were only 18. 

This has good explanations about the functionality about to be executed, then the output is produced and there is no explanation of how someone should interpret the output.  Examples ( 2 of many places)

-  I got down to the end of section 5 and you said, "If the mean and variance is somewhat similar then we consider that seasonality has been removed from the series" ... and as I looked at the 2 columns, I couldn't reconcile the statement made before the SELECT with the MEAN and SD columns. 
- At the end of 7, you said "Creating a table which will store the results for the ts_sf_estimated_series_plot build using TD_Plot" but when I opened the graph, all I saw were red saw-toothed lines and blue saw-toothed lines and don't know what the relative size or the correlation of the peaks mean. 
- The image at the end of 10 also needs explanation

I would examine every output and pretend you are going to answer my question, "What does this mean?"  If you can't explain it, ask yourself if the table of numbers contributes to the business story you are trying to tell. 

Long running statements:
runsecs	querytext
  176.080000	"SELECT * FROM DEMO_UAF.nn5_data ORDER BY DateTime sample 5;"  ... forces entire table to be processed

PR #266, opened 2023-05-16
Car

nupur-lal:
New usecase notebook on car complaints analysis


DougEbel:
Although merged, there is still performance work to be done.  One step is doing a product join for the geospatial and needs to join using a MBR as mentioned on e-mail of 5/11.

PR #268, opened 2023-05-16
Converted sql to py-sql version and added map to visualize bike stations

chetan-hirapara:
As Teradata-kernel we can't use Python libraries for visualization. So I have converted this notebook from **SQL to py_sql version** and then **added maps to visualize the stations**.

- Here issue: https://github.com/Teradata/jupyter-demos/issues/168

@DougEbel @mrconway Please review the changes made to make this notebook work. Thanks.


DougEbel:
GREAT IMPROVEMENTS!! ... however a bit of touch-up needed before merging:
1. Notebooks should be reset and cleared of outputs before saving. Otherwise, users can't tell that anything is being run ... however someday we would like to have a PDF (or other) for pure browsing and looking at the pictures. 
2. you commented out !pip install folium, but we need to add it to the image.  You can comment it out, but see the standard text GDC has been following the commented out pip installs. 
3. Need to hide exceptions like:  eng.execute("drop table weather_temporal")
4. You are pretty good at ensuring that there is an output after graphs to say what the user should notice about the above output.  There are some table outputs that lacked that description ... e.g. tdf_trips_geo_pti
5. you might describe temporal qualifiers first time each is used like "SEQUENCED VALID TIME" and VALIDTIME PERIOD (...) and benefit vs raw SQL ... perhaps interaction with the PTI.

DougEbel:
Running Teradata/Support_Macros.ipynb , and then using: 
exec recent(4)    ... last 4 hours
exec qrylog(1287)   .... passing the sessionid of the session for Austin bike share:

the longest queries were:
runsecs	querytext
   40.150000	...INSERT INTO trips_geo_pti
   37.380000	... CREATE TABLE trips_and_weather AS (
   
probably not much you can do but you might add some description of what it is doing and why it might be slow based on the number of rows or columns or calculation complexity on a 2-cpu platforrm.

DougEbel:
Standard commented pip install:

Installing some dependencies

# %%capture
# # '%%capture' suppresses the display of installation steps of the following packages
# !pip install sklearn2pmml --user
# !pip install jdk4py --user
The above statements may need to be uncommented if you run the notebooks on a platform other than ClearScape Analytics Experience that does not have the libraries installed. If you uncomment those installs, be sure to restart the kernel after executing those lines.

chetan-hirapara:
> Standard commented pip install:
> 
> Installing some dependencies
> 
> # %%capture
> # # '%%capture' suppresses the display of installation steps of the following packages
> # !pip install sklearn2pmml --user
> # !pip install jdk4py --user
> The above statements may need to be uncommented if you run the notebooks on a platform other than ClearScape Analytics Experience that does not have the libraries installed. If you uncomment those installs, be sure to restart the kernel after executing those lines.
-------------
* Done

chetan-hirapara:
> Running Teradata/Support_Macros.ipynb , and then using: exec recent(4) ... last 4 hours exec qrylog(1287) .... passing the sessionid of the session for Austin bike share:
> 
> the longest queries were: runsecs querytext 40.150000 ...INSERT INTO trips_geo_pti 37.380000 ... CREATE TABLE trips_and_weather AS (
> 
> probably not much you can do but you might add some description of what it is doing and why it might be slow based on the number of rows or columns or calculation complexity on a 2-cpu platforrm.

--------------
* Added description about query data and joins.

PR #271, opened 2023-05-17
train delay incorporated review comments

nupur-lal:
Adjusted some parameters in model creation and added explanation for F-STAT &  F-conclusion


DougEbel:
This still shows "Abnormal Wheater conditions" in the data.  That could be fixed without changing the notebook.

PR #278, opened 2023-05-24
Demand forecast

shilpa-nalkande:
Updated review comments as per mail sent and waiting for response from Denis on some points

### DIFF
--- a/UseCases/DemandForecasting/Demand_Forecasting_Python.ipynb
+++ b/UseCases/DemandForecasting/Demand_Forecasting_Python.ipynb
@@ -1,0 +1,856 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "07c7b2ca-a3a2-4525-aeb3-2276de42e3c2",
+   "metadata": {},
+   "source": [
+    "<header style=\"padding:1px;background:#f9f9f9;border-top:3px solid #00b2b1\"><img id=\"Teradata-logo\" src=\"https://www.teradata.com/Teradata/Images/Rebrand/Teradata_logo-two_color.png\" alt=\"Teradata\" width=\"220\" align=\"right\" />\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


DougEbel:
This is a pretty good demo.  I want to take another look in the AM.  Initial reaction is questioning the data of 6 stores, showing a bar graph where 2 of the markets have almost no volume ... and the graph with top 250 products.  That would be ok if you spin it that trying to predict demand using simplistic graphs is confusing.

DougEbel:
1. In intro, drop paragraph  or combine with paragraph 3.   Then check spacing after paragraph 3
2. State that you are omitting market01 and market06 because there is insufficient data.  Drop table, only need bar chart. 
3. Need to parenthetically describe duration and MATERIAL in the statement: "We will check only those Series where the duration (...) is greater than 50 and MATERIAL (...) less than 300."
4. need to put statement after the top 250 graph that traditional charting for of demand is a confusing mess of spaghetti  (or something like that)
5. One of our strengths is that we can run thousands of models concurrently, however for the purposes of this demo, we will arbitrarily choose 3 products so you can follow the process.
6. I'm puzzled that MARKET01 shows up on "dataset_new" when it had almost no business.  If that gets filtered out later, then say so.
7. Define stationary on first use
8. Missing explanation of the PACF_out.result
9. I've seen the graphs like the 2  Autocorrelation and never understood what the blue shaded area was.  I see that "ROW_I is the X axis"  ... don't know what ROW_I means.
10. missing description for results_estimate
11. missing description (brief words ok) for arima_validate_op.result
12. In your 3 graphs of calc value vs demand, not sure I understand the orange vs blue.  For the first and third set of orange dots, they seem to follow the green line but the middle one is not a very good orange plot.  However the blue dots are very close for all 3. 
I'm going to merge this one because it is functional but could use some improvements and it is easier to see the changes for the next pull request.

PR #3, opened 2022-08-19
Added Time Series Forecasting Demo

ForbiddenDevil:
@DougEbel Please review the TimeSeriesForecasting Demo.

PR #279, opened 2023-05-24
predict Remaining useful life for engines

chetan-hirapara:
Ticket: https://github.com/Teradata/jupyter-demos/issues/221

@DougEbel Please review the changes made to make this notebook work. Thanks.


DougEbel (commented):
1. You failed to do restart/clear all outputs.  Notebook was 4.9 GB and with results from your execution, the user doesn't know that anything is being executed. We could just be posting PDF's
2. I wonder but not sure it the display of the source data would be better sorted.
3. you list out the df columns without comment
4. I ran all steps automatically, in step 7 for df_eng_cycles.head(10), it create and error: error was hidden in pages of junk:
OperationalError: (teradatasql.OperationalError) [Version 17.20.0.0] [Session 1393] [Teradata Database] [Error 3807] Object 'predictive_maintenance_rul' does not exist. ..

I  can't merge this ... I"ll look tomorrow to see if there is a simple fix.

DougEbel:
The failure was due to missing the database name on 2 select statements. ... Continuing the review.

5. you need to spell check the document, look for correct plural usage,  Examples:

- Now, Let's do some data **exploraion** with engine number and time in cycle variables.
- Although these columns **has** 100% null values

6. Before saying "Please scroll down ...", provide a generalized description of the data: The data is from 100 engines with hundreds of rows of measurements collected from each time the engine was cycled.  There are three operational settings that have a substantial effect on engine performance. Data is collected from 26 sensors.  There is a row for every time the engine is cycled (starting at 1)  and a column counting down the cycles until maintenance was required labeled RUL or Remaining Useful Life.  
7. I don't think this in #3 is worded consistently with demo:  Let's look at the sample data in the trips table
8. I think the "Time in Cycle by Engine number" chart may be more accurately described as number cycles until maintenance needed. ... after the graph, I don't thing "has time in cycle"  or "average number of time in cycle" is consistent with the data. 
9. In the very interesting "In the above pairplot, we observe a few things"  ... I don't know how you got those observations.  Add an explanation after...because ... on lines 1 and 2. 
10. When you get to XGBoost, the prediction,, confidence_lower and confidence_upper are all numbers in the ~97.nnnn range.  I'm no longer recognizing what is actual and what is predicted and what is the engine. 
11. in several places you put the words "The above graph displays the ..." but above is a table.  And it has the F_Conclusion needs words to explain ... Non-informed user might think the prediction failed. 
12. in the "Actual vs Predicted using DecisionForest Classification" graph, the x-asis says Engine ID, but  the **engine_no** has a range of 1 to 100.   The lines are so dense, it is difficult to tell visually how close the estimate and actual are.  Maybe just show  the RUL estimate for 1 engine or a few engines, or one line for the engine and one for difference between estimated and actual. 

... This is a great demo with interesting graphics!!

chetan-hirapara:
> 1. You failed to do restart/clear all outputs.  Notebook was 4.9 GB and with results from your execution, the user doesn't know that anything is being executed. We could just be posting PDF's
> 2. I wonder but not sure it the display of the source data would be better sorted.
> 3. you list out the df columns without comment
> 4. I ran all steps automatically, in step 7 for df_eng_cycles.head(10), it create and error: error was hidden in pages of junk:
>    OperationalError: (teradatasql.OperationalError) [Version 17.20.0.0] [Session 1393] [Teradata Database] [Error 3807] Object 'predictive_maintenance_rul' does not exist. ..
> 
> I can't merge this ... I"ll look tomorrow to see if there is a simple fix.

* changes done

chetan-hirapara:
> The failure was due to missing the database name on 2 select statements. ... Continuing the review.
> 
> 5. you need to spell check the document, look for correct plural usage,  Examples:
> 
> * Now, Let's do some data **exploraion** with engine number and time in cycle variables.
> * Although these columns **has** 100% null values
> 
> 6. Before saying "Please scroll down ...", provide a generalized description of the data: The data is from 100 engines with hundreds of rows of measurements collected from each time the engine was cycled.  There are three operational settings that have a substantial effect on engine performance. Data is collected from 26 sensors.  There is a row for every time the engine is cycled (starting at 1)  and a column counting down the cycles until maintenance was required labeled RUL or Remaining Useful Life.
> 7. I don't think this in [Added Time Series Forecasting Demo #3](https://github.com/Teradata/jupyter-demos/pull/3) is worded consistently with demo:  Let's look at the sample data in the trips table
> 8. I think the "Time in Cycle by Engine number" chart may be more accurately described as number cycles until maintenance needed. ... after the graph, I don't thing "has time in cycle"  or "average number of time in cycle" is consistent with the data.
> 9. In the very interesting "In the above pairplot, we observe a few things"  ... I don't know how you got those observations.  Add an explanation after...because ... on lines 1 and 2.
> 10. When you get to XGBoost, the prediction,, confidence_lower and confidence_upper are all numbers in the ~97.nnnn range.  I'm no longer recognizing what is actual and what is predicted and what is the engine.
> 11. in several places you put the words "The above graph displays the ..." but above is a table.  And it has the F_Conclusion needs words to explain ... Non-informed user might think the prediction failed.
> 12. in the "Actual vs Predicted using DecisionForest Classification" graph, the x-asis says Engine ID, but  the **engine_no** has a range of 1 to 100.   The lines are so dense, it is difficult to tell visually how close the estimate and actual are.  Maybe just show  the RUL estimate for 1 engine or a few engines, or one line for the engine and one for difference between estimated and actual.
> 
> ... This is a great demo with interesting graphics!!

* Done
* Point 10: renamed column RUL to actual. There are id field is unique so adding it to prediction result set. 
* Point 11: Added description of F_conclusion
* Point 12: Taking the first 50 records to the visualization for better interpretability. Also, removed title engine no.

chetan-hirapara:
@DougEbel Please review it again and let me know for more details.

PR #280, opened 2023-05-25
Behavior analysis

nupur-lal:
@DougEbel @chetan-hirapara 
from the comments on PR#272, BehavioralAnalysis_Python.ipynb has full python code and BehavioralAnalysis_PY_SQL.ipynb has code for sessionize and npath using sqls and visualizations in python.
* updated npath visualizations in both the notebooks 
* updated the bar plot from original notebook
* updated markups

### DIFF
--- a/UseCases/CustomerBehaviorAnalysis/BehavioralAnalysis_PY_SQL.ipynb
+++ b/UseCases/CustomerBehaviorAnalysis/BehavioralAnalysis_PY_SQL.ipynb
@@ -41,24 +41,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "<p style = 'font-size:16px;font-family:Arial'>Installing python libraries which are not installed by default in the environment.</p>\n"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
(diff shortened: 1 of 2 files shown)


chetan-hirapara:
Overall changes look good, however, few small points to check:

1.  There are inconsistent spaces in `import libraries`  
2. Use `db_drop_table` function in `PY_SQL.ipynb`  instead of `eng.execute('DROP TABLE DEMO_USER.demo_sessionized_events;') `
3.  It would be more readable if you can divide it into 2-3 paragraphs. 

> To check the details of any path or node we can move the mouse pointer over it and check details. For example if you move the pointer over the dark Green path having the largest width at the TOP going towards the right most node(Purchase).The number/count shows there number of entities who followed that path starting from Product Return ---> Purchase.
When the pointer is moved over a Node, for example when the pointer is on the long purple Node at the right top Product Return it shows incoming flow count and outgoing flow count. Incoming flow count means the number of different event which led to the event in consideration and outgoing flow count the number of different event after this event. Similarly other nodes and paths can be analyzed.

DougEbel:
This modified notebooks under old directory/notebook names.  Files were committed using a new branch: Customer_Behavior_Analysis and the branch BehaviorAnalysis was deleted.

PR #288, opened 2023-05-29
New notebook for MultiTouch Attribution

shilpa-nalkande:
Created as per the notebook provided. Will continue working and compare with Tableau charts and other Details . Also let me know your comments so that I can include all changes in next pull request

### DIFF
--- a/UseCases/MultiTouch_Attribution/MultiTouch_Attribution_PY_SQL.ipynb
+++ b/UseCases/MultiTouch_Attribution/MultiTouch_Attribution_PY_SQL.ipynb
@@ -1,0 +1,3115 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "5169c9ee-1e13-461e-b599-d615c9873035",
+   "metadata": {},
+   "source": [
+    "<header style=\"padding:1px;background:#f9f9f9;border-top:3px solid #00b2b1\"><img id=\"Teradata-logo\" src=\"https://www.teradata.com/Teradata/Images/Rebrand/Teradata_logo-two_color.png\" alt=\"Teradata\" width=\"220\" align=\"right\" />\n",
+    "\n",
+    "<b style = 'font-size:28px;font-family:Arial;color:#E37C4D'>Multi Touch Attribution using Vantage</b>\n",
+    "\n",
+    "<p style = 'font-size:20px;font-family:Arial;color:#E37C4D'><b>Introduction</b></p>\n",
+    "\n",
+    "<p style = 'font-size:16px;font-family:Arial'>Marketing attribution modelling techniques aim to determine the contribution of each marketing touchpoint or channel in influencing customer behaviour and driving conversions. These models provide valuable insights into the effectiveness of marketing efforts, helping businesses make informed decisions regarding resource allocation and optimization.</p>\n",
+    "<p style = 'font-size:16px;font-family:Arial'><a href='#rule'>Rule-based</a> attribution modeling relies on predetermined rules or heuristics to assign credit to various touchpoints along the customer journey. Common rule-based models include the First Touch, Last Touch, Uniform (linear) and Exponential(time decay) models. The First Touch model attributes all credit to the first touchpoint a customer interacts with, while the Last Touch model assigns all credit to the final touchpoint before conversion. The Uniform model evenly distributes credit across all touchpoints in the customer journey. The Exponential model assigns more credit to touchpoints closer to the conversion event.<p>\n",
+    "    \n",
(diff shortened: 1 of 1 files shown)


chetan-hirapara:
@shilpa-nalkande @DougEbel - I have reviewed this notebook, Overall it looks good. You have added sufficient visualization too. 

Here are my review points, that would enhance the explanation and usability of this demo.

• It would be great if you can add one multitouch attribution graphic like this: 

-  https://cdn-www.oktopost.com/blog/wp-content/uploads/2020/10/01-digitalmarketingcampaigns-02.png
-  https://staging.wordstreamcdn.com/wp-content/uploads/2021/12/types-of-attribution-models-google-ads.png
-  https://us.principle-c.com/wp-content/uploads/2020/01/mta-sales-nielsen.jpg

		
•  Links are not working in Introduction i.e. Rule-based, Statistical etc
• Please describe the data first in section 3. Add a data source, and features, what is this data about?
• It would be great if you can add a chart to visualize channel_df (bar chart or pie chart)
• Add explanation to Sankey Diagram chart - what insights are drawn from this?
• Please description to these points:

-    Uniform
-    First Click
-    Last Click
-    Exponential

• Add a little bit of description to all the code blocks in section 5.2 like Uniform,  First Click, etc.
• I would suggest adding one function for the bar chart and calling it with the required parameters to gain advance of core reusability.  
• Please add CorrMatrix to  chart(heatmap) for better visualization.
• Add a little bit of description on td_analyze and which rescaling approach you're using
• In the RF features importance graph I title is wrong "Random Forest Model Attribution Summary". It features importance graph. Also, sort it in descending order before display.

DougEbel:
If there is an output that is not followed by an explanation, then add it.  For example:
- what do I conclude about the spaghetti in the path diagram?

in 4.2 you are using the npath as data preparation.  How?  what was used? If you are creating tables to be used in multiple functions, then outline the functions that will use the data.  You have converting and non-converting journeys in the 2 data sets.  Explain how they will be used.  (all of this in a few sentences unless there is a bulleted list of where data is used).

In 5. .. agree with Chetain on putting a definition next to the rule based models

Prior to 5.1 you say you will consider 20 rows from most recent ... but then in in 5.1 code, there is no reference to "20" rows. 

You don't describe how TABLE ATTRIBUTION_MODEL_FIRSTCLICK is being used. How does that relate to "20" rows?   
... you might be assuming that everyone is an expert programmer and deeply studying and understanding each and every query.  My first pass is looking at the words in the markdown cells and then there is a bunch of code which bears no resemblance to  the markdown description.  So I get confused (on behalf of the public who will be viewing this in the future).

In the 5.4 graph, you show that all 4 approaches agree that Facebook is the best, Part of my initial frustration with the demo is that being 1/3rd of the way through the steps, you are going to tell me over and over that facebook is the best.  Why are you wasting my time?  The answer is that there may be other factors to consider and other approaches that will yield a better analysis (I hope). 

Assume the reader is skimming.  when you make a statement like in 5.5 as "marketers can gain insights into the additional value or improvement offered by more sophisticated approaches." ... follow with " ,, such as First click, last click and exponential which follow."

In the graph at the end of 5.5, it would be good to remind the "skimmer' what the attribution score means which is the "X" axis.  The casual reader might float over (not mentioned) the light blue ball which is bigger and think that is better but it has a higher cost than the dark blue ball and a lower "total attribution" (? meaning?).

Is this the only consideration?   Online display has a cost of 13.445, online video has a cost of 88,975 with a bigger bubble than Facebook with a cost of 67.305.  Perhaps an additional view switching up time to conversion with cost (only because I don't understand the Total attribution dimension). 

Not sure I understand the comment in 6.1: "(here paths) into "terms" (grams) of selected size (1) and count them." .  What do you mean by a "Simple frequency analysis"?  is that the number of times that an activity appeared in the path to conversion?

Now that I'm down at 7.1, I'm wondering if I'm still processing 20 rows. 

In 7.1, how does this analysis differ from the statistical analysis that gave an attribution weight of about .33 with the association analysis that gives an attribution weight of .45? 

Just looking at the headings, I'm confused about the difference between: 
7. ASSOCIATION ANALYSIS
   Association analysis can assist in implementing multi-touch attribution models. ...
8. ASSOCIATION ANALYSIS
   Association analysis can help identify channels that are frequently used in combination with other successful channels. 

.... I am stopping the analysis of the notebook here.  Chetan has added additional comments on the content below here.  I am going to merge the notebook so it is available in the folder view of Jupyter for associates interested in MTA so they can review it and send me feedback, but won't be adding it to the Readme.md index until this weekend or next depending on improvements to the content.  Merging also makes it easier to review the improvements in github.

PR #290, opened 2023-06-01
Energy Consumption Forecasting Dataiku

ForbiddenDevil:
@DougEbel I've added Dataiku steps in the demo. Please have a look. Thanks.


DougEbel:
I got to the step: "# Obtain a pointer to the model" and got the error: "TeradataMlException: [Teradata][teradataml](TDML_2204) Failed to retrieve the model. Table "demo_user.dataiku_models" does not exist"

I changed to the branch and ran all steps automatically using the >> Icon at the top. 

The demo is looking good so far

ForbiddenDevil:
In this demo, the models and train/test data were originally created using the Dataiku platform. However, if you need to run the demo without access to Dataiku, you will require the model and data files. Therefore, I have included these files and added code to set up the necessary environment without relying on Dataiku.

DougEbel:
1. You link the data but don't set up the relationship for this demo to real usage.  You might need a diagram that explains you are getting data from the cloud so that you could run Dataiku and do the modeling creating a PMML file which is imported for scoring. Alternatively (as the demo shows, you will include screen shots of what would be happening in DataIku and the pick up the demo at the end by importing the PMML. 

2. not clear what "2. Dataiku (In-Progress)" means

3. I wonder if the python code should be in Jupyter or the recipe.  I suppose ok in python, but need to be clear that DataIku is running stuff instead of Jupyter.

4. In the only visualization, the charts is so dense that I can't follow the assumption that Random Forest is better than RMS.  How about a supplemental  chart. 

I'm going to close the pull request.  The changes should be pretty quick.

DougEbel:
I added a paragraph to the intro explaining they don't need Dataiku to run the demo before I merged it.

PR #291, opened 2023-06-01
Demand forecast

shilpa-nalkande:
Added pip install and comments from Denis. Updated based on review comments.


DougEbel:
This demo is merged.  

However, I'd like someone to explain the "5.3. Plot graphs for ACF and PACF of the time series"  ... what is the blue background on the 2 graphs?   Should we keep that a secret from the public?

PR #292, opened 2023-06-02
Updated Review comments for MultiTouch Attribution

shilpa-nalkande:
Updated review comments. Please add the images folder also when merging to get the image


DougEbel:
1) It was requested for the nPath to have a single end point - conversion. Why are there 5 conversions?

2) The words said the 2nd Sankey was based on total cost, but there was no explanation (unless I read and understand the nPath call

I went ahead and merged.  There may be additional updates.  I believe the pip installs are no longer needed

PR #302, opened 2023-06-08
Hyper personalization

nupur-lal:
@DougEbel updated and renamed the notebook as per the comments given.
Changes:
Intro description as per mail
new chart for incomeband vs churn
changes in dataprep wrt to incomeband and to create segments
model creation and scoring and comparison are done via function to loop it for all the segments


DougEbel:
This is pretty good, so I've merged it.   

In the (long) introduction, you have a single <br> between paragraphs.  Consider both the consistency of format and the length of the introduction. 

You might want to summarize the "Data Exploration" steps you will follow.  Some of the patterns are fascinating., and then (in the introduction) you are going to do some transformations ... because?

I got a bit lost in "Model Training and Scoring for each segment" and didn't know how to interpret the plethora of graphs.  You might want to provide some explanation before and after the set of graphs.

I'm merging the pull request because it is functional (but could use some improvements)

PR #310, opened 2023-06-13
Uaf cashdemand

nupur-lal:
changes from the original notebook from Tim/Paul
* in sql kernel instead of python as all the code was in sql
*python boxplot was removed due to sql kernel
*td_diff and td_seasonalise were removed as dikey_fuller was giving series as stationary due to recent updates in function
* model parameters were changed acc or acf & pacf plots
* removed TD_DurbinWatson & TD_BreuschGodfrey  as they were only for demonstration not useful in demo


DougEbel:
1. paragraphs in the intro should have a blank line between
2. Very good descriptions down to 4- Line plot: Applying the TD_Plot function and building an ART table for the results ... a bunch of zig-zags and dots.   If the purpose is to say that more analytics are needed because simple line graphs don't reveal patterns, then ok. But don't leave noise like that in demos without explanations. 
3. You have 63 references for "NN5_3" and I don't see a definition.  Might be useful if you tied that to the business problem somehow.  I'm guessing it is some cash demand for 1 of  111 ATMs and you chose one because _____?  possibly to allow the user to focus on one and eliminate the confusion of seeing 111 graphs. 
4. The scatter plot ... Description please?  If you can't describe what it means, don't include it.  Save the demonstration of all of the things tdplot can do for a demo on tdbplot or a demo where the dots mean something.
5. When I look at the line graph, it looks the beginning and ending month of the 3 years of data (month 3 of 1996 and 1998) should have been an outlier because there was little activity. 
6. If you say "printing 10 rows of table"  then issue:  select * from ACF_RES_tw; and get 15 rows, I suspect something is wrong
7. If you say "Printing some rows of the table" then issue: select * from PACF_RES_tw;,  I suspect something is wrong
8. If you say "TD_Plot for stationary panel makes a 2 x 2 window each with a plot you made above like ACF, PACF, Line plot. It is build to get the insights ", produce the graph and don't provide explanations of what insights  should have gotten, then you haven't finished your work on the demo. 

I am not half way through the demo and have spent an hour trying to follow it.  There is lots of good content but missing some explanations.  It is long. Feel free to eliminate cells that are weak contributors to supporting the business issue.  Don't feel like you need to show every thing that can be done with every function.  There will be more demos to show them.  It is fine to list in comments (e.g.) the types of graphs that can be produced.  then say, for this demo, xxxx will be used to show yyyyy.    Show the output, then explain what it means.

DougEbel:
Using the macros recent(4) and qrylog, I found that the getting data for local storage took 619 seconds.  Of that, 563 seconds was 
" insert into DEMO_UAF_db.Nn5_Data select * from DEMO_UAF.Nn5_Data;".  Is this really a demo of UAF when you are going to only process 1 column out of a bloated array?   

Feel free to discuss with others because I may not appreciate the value of UAF in this case. If it were me, I'd transform the raw data into a 3 column table (date, ATM and Amount).  

I'm not sure where the time is being spent on the insert/select.  One diagnostic is to do a count(*) on the foreign table where ... texting each of the 111 column being >= 0.  That forces all of the columns to be converted out of parquet and considered as a row. 

With about 2 years of data you've got about 700 UAF rows.  Is that what caused the narrow zig-zag lines in the first graph?  Wouldn't it be more interesting to first look aggregated to the week or month level, then look at aggregate by day of week?  (I assume there are more withdrawals on Friday and Saturday.

PR #312, opened 2023-06-14
Uaf cashdemand

nupur-lal:


DougEbel:
1. "SELECT TOP 5 * FROM DEMO_UAF.atm_data;" worked but "Select count(*) from atm_data;" said object atm_data doesn't exist.  Adding database reference so I can continue. 

2. On first cell of "2. Data Cleaning", get error message again: "Unable to run SQL: Unable to run SQL query: Database reported error:3807:Object 'atm_data' does not exist."

... stopping review, closing pull request.

PR #316, opened 2023-06-19
Modelops Beta message to modelops_v0 branch

pabloesc:
00 - adding beta message
02 - removing links after credits
10 - removing links after credits


DougEbel:
I'll refine the comment in the beta note.  They will need to provision another machine without modelops.  We do want them to let us know about problems.

PR #319, opened 2023-06-23
New SalesForecasting Demo using STO

shilpa-nalkande:
Please review and let me know changes to be done

### DIFF
--- a/UseCases/Forecasting_using_STO/StoreSales_Forecasting_ProphetSTO.ipynb
+++ b/UseCases/Forecasting_using_STO/StoreSales_Forecasting_ProphetSTO.ipynb
@@ -1,0 +1,1101 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<header style=\"padding:1px;background:#f9f9f9;border-top:3px solid #00b2b1\"><img id=\"Teradata-logo\" src=\"https://www.teradata.com/Teradata/Images/Rebrand/Teradata_logo-two_color.png\" alt=\"Teradata\" width=\"220\" align=\"right\" />\n",
+    "\n",
+    "<b style = 'font-size:28px;font-family:Arial;color:#E37C4D'>Store Sales Forecasting with Prophet using Script Table Operator </b>\n",
+    "</header>\n",
+    "\n",
+    "<p style = 'font-size:20px;font-family:Arial;color:#E37C4D'><b>Introduction</b></p>\n",
+    "\n",
+    "<p style = 'font-size:16px;font-family:Arial'>Rossmann operates over 3,000 drug stores in 7 European countries. Currently, Rossmann store managers are tasked with predicting their daily sales for up to six weeks in advance. Store sales are influenced by many factors, including promotions, competition, school and state holidays, seasonality, and locality. With thousands of individual managers predicting sales based on their unique circumstances, the accuracy of results can be quite varied.<br>\n",
+    "<br>  \n",
+    "Our Main Objective is to predict sales of store in a week. We are using the python Prophet model and using the Script Table operator of Vantage for forecasting the Store Sales.\n",
+    "<br>\n",
+    "    \n",
+    "<p style = 'font-size:16px;font-family:Arial'><b>STO is a convenient and useful way to use external algorithms along with native Vantage functions to operationalize insights WITHOUT needing to sample data or create data silos for analytics.</b></p>\n",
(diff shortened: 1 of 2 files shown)


DougEbel:
Nice demo. Used "exec recent()" and "exec qrylog(nnnn)" to assess performance and only the STO had a lengthy response of 47 seconds. You might comment on that before executing.  I like the fact you chose some sample stores to manage performance.

details:
1. There is no suffix on the notebook name indicating type. This appears to be mostly a Python demo.  I seem to recall there were issues with Python doing STO, so it is ok to use the Python  suffix.  If someone comes up with a pure python version, we can do renames to support both versions. 
2. On first reference to STO abbreviation, use the full words followed by (STO) .  Literary practice is to bold/underline the letters in the word used to derive the abbreviation.  That should be in paragraph 2 of the intro
3. In the intro, paragraph 3 is randomly in bold.  I would combine/rewrite paragraphs 3 and 4 to make it clearer what STO is doing:   STO allows you to leverage the parallelism of Vantage by installing external routines into each unit of parallelism.  During the execution of the analysis, data is piped to and from those external routines in dozens, hundreds or thousands units of parallelism depending on the size of the production platform.  This allows you the flexibility and power to extend the analytic power of Vantage with routines written in languages such as Python or R to process data in parallel without exporting to another platform.   .... new paragraph ... The SCRIPT function enables invoking ... and the rest of paragraph 4.
4. I would add an image in the intro.   Here is an example:
[Sample](https://teradata-my.sharepoint.com/:i:/p/douglas_ebel/EcMRoxu2rHxDo8lAEBi5CzQB87vVDKFVMAmq-9EwahwXgw?e=9Keh9d)
the html:   
5. on 3. Prepare data, indicate you are creating data frames for store and sales and showing 5 sample rows. 
6. In the seaborn 2x4 set of graphs,  remind the reader that the Rossman company had 4 types of stores and in total there were 3000 stores.  Each dot in the graphs above represent the volume for 1 store for one month. ... or is it multiple years aggregated to month?
7. You have " ... to represent their group**. "  which indicates there is a footnote somewhere but there isn't. 
8. You seem to be claiming that all store type D's are closed July-Jan or did that just happen for the 1 you chose to demonstrate.   If they all were not closed, then choose a different example for "D" because this is unnecessary noise in your story. 
9. I'm confused about the yearly trend.  First, you are using seasonal_decompose without saying what it does.  there are lots of spikes and dips in the graphs is that weekly?   The 2nd graph has a different x-axis scale than the other 3.  Somehow you seem to be able to determine that overall sales increase but looking at the squiggles, I don't understand how you arrived at that conclusion. 
10. In 5, checking for Stationarity ... you say that Dickey fuller (capitalization?) needs series data ... meaning?  You've had 4 store types but there are rows 0 - 4 output .... what is the 5th row? 0 and 4 say then "NULL_HYP" is ACCEPT but your statement following the table says "... in all cases we consider it as stationary"
11. In 6. Creating the Model ... you say "We then use this py file in the STO command" .... isn't it the SCRIPT function?
12. I'm trying to understand if Prophet is doing something that is separate able from STO, meaning section 6 should be broken up into 2 sections with a description for each.  It is not clear on a quick scan of the stoSalesForecastnew.py what it is doing.  There should be a general description of what it does, then in the stoSalesForecastnew.py file, there should be a comment block at the top that describes what the program does and describes its input and output. 
13. The following looks like a spelling error " ... found:  includes a column yhat with the forecast," .  Instead say "includes the "yhat" column ..."
14. When you show "STO_out.shape state that it has 1004 rows, one for every ___ and 7 columns of information ... followed by the bullets. 
15. Then you start doing some transformations prior to the Forcast Sales Values with Confidence Interval graph ... explain what you are doing
16. On that graph, the confidence intervals go negative.  Am I to interpret  that some of the range of of prediction can go negative?  If I go back to the sample of 5 rows, 47	2012-12-05	7584.492554	600.882268	13918.175312 ... which is an estimate of 7584 with a possible range of 600 to 13918.  Really?  How are they make any business decisions based on such a range?   If there is a different interpretation of the confidence level, include it. 
17. For the 3 plots, you say "The first plot shows that the sales has been linearly decreasing over time."  The graph is scary, however if you think about it, dropping from 4687.5 to 4670, a drop of 17.5 or  0.37% which while not growing, it is essentially flat.  On the second graph, how can the weekly volume of sales be similar while the third graph shows yearly volume varies ... interestingly down to negative -2000  .... are those returns?  (ok ... silly comment), but I'm not understanding the graphs and associated explanation. 

Its not a blocker for publishing this demo, but I wonder if the data scientist will be frustrated that you don't build a forecast on the first 80% of the time periods and then show the accuracy vs the last 20%.   Is the forecasting done on the sum of the 4 stores?  Would there be less range of confidence if storetype D was excluded ... or separate forecasts instead of the sum.

DougEbel:
The html for the graph got interpreted instead of shown.  It was:

img src="STO.png" align="right" width="400"

PR #321, opened 2023-06-27
Demand sto

shilpa-nalkande:
Updated review comments and changed Name for Forecasting demo using STO


DougEbel (approved):
I changed the image to be right justified and smaller.

PR #322, opened 2023-06-27
incorporating the review comments

nupur-lal:
incorporated review comments
* updated introduction as per Michael Senger 
* note for sql alchemy output without  any display
* changes in model creation and scoring segment
* typos

### DIFF
--- a/UseCases/Hyper_Personalization/Hyper_Personalization_PY_SQL.ipynb
+++ b/UseCases/Hyper_Personalization/Hyper_Personalization_PY_SQL.ipynb
@@ -19,22 +19,17 @@
     "<p style = 'font-size:18px;font-family:Arial;color:#E37C4D'><b>Introduction</b></p>\n",
     "\n",
     "<p style = 'font-size:16px;font-family:Arial'>\n",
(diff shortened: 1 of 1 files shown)


DougEbel (approved):
Sorry for not referring to section number ... there aren't any.

You say "the above chart shows a heat map" but you don't say what the values mean or point out a few interesting correlations. 

what is Monthly Chareg  ?

When you create segments, you probably should begin with a bulleted list of the segments you will create.  You say that it was based on the profiling, so above when you are profiling, make a reference to the bulleted list that will follow, like "One of the segments we will want to analyze is SeniorCItizen"

You need a major section head for Defining the functions.   You might start out the section saying we will define functions to ____ that will will call in the next section to perform ____ ... because in scanning the text, I saw "plt.show() and was about to complain there was no plot produced.  The specific groups of functions then get minor section headings, and the next section has a title that you are Performing model training and scoring.  The first sentence in that major section says, we will now use the functions defined above to loop through the segments: ... and then bullet list them so the reader is reminded. 

You have a long string of graphs that are hard to read (at skim speed) and understand.   What if you separate out the definition of the segment list in its on cell, and describe that is the list of segments you will be processing.  Then describe you will loop through the 6 segments. In your loop, print a blank line and a banner like:  "====== SenionCitizen =====".  Then following the set of graphs, in a bulleted list with 6 bullets, describe your interpretation of each segment.   

In your conclusion, you should say that this is a simple example of hyper personalization, At its extreme (I think?) we want to define smaller and smaller segments so we can give more personalized attention to each and every customer.    (Ask some of your team if adding that statement is extreme). 

I'm going to merge this because it is easier to review and it doesn't have any major failures, but the usability/understandability could use some improvement (and spell check).

PR #338, opened 2023-07-07
Anomaly Detection: Fixed ROC error

ForbiddenDevil:
@DougEbel This notebook threw error while running the ROC function. Fixed the error. Please review and merge. Thank you.


DougEbel (approved):
Change looks good.  Pranay hit the error before I got it merged, but I just tested before and after the git switch and the fix is good.   Question:  It worked before when I merged it.  Why did it fail now?

Issue #335, opened 2023-07-06
Generative Question Answer using GenAI


chetan-hirapara:
Create a Generative Question-Answering Demo using LLM Models

We need to develop a Generative Question-Answering demo that utilizes Large Language Models (LLM) to provide accurate and natural language responses. The goal of this demo is to allow users to ask queries about Vantage tables using layman's terms and receive appropriate answers generated by the application.

PR #339, opened 2023-07-07
added new notebook for openai api key

chetan-hirapara:
@DougEbel - Please review it 

Task: https://github.com/Teradata/jupyter-demos/issues/337

Please note: This is initial version, I have created it in some urgency to start your work to create api key. Later we'll modify it

PR #340, opened 2023-07-10
Generative Question Answering using Generative AI with Vantage

chetan-hirapara:
Hi @DougEbel and @ForbiddenDevil - Please review this PR.

**Task description:** 
* Create a Generative Question-Answering Demo using LLM Models

*  We need to develop a Generative Question-Answering demo that utilizes Large Language Models (LLM) to provide accurate and natural language responses. The goal of this demo is to allow users to ask queries about Vantage tables using layman's terms and receive appropriate answers generated by the application.

**Task url :** https://github.com/Teradata/jupyter-demos/issues/335


UseCases/Generative_Question_Answering_GenAI/images/vantage_qa_gen.png
DougEbel, line 1: Until the modules are merged into the image, the !pip install needs to be uncommented.  I'll change it when (if) installed into the image.  I like the requirements file approach.  Too bad we didn't do that earlier.

suggest 0 0 for restarting the kernel

After step 1, need bold step 1.1 or 2 that says to get the api key and then insert it into a variable or prompt for it (like we do for password .  That needs to happen after the 0 0 to reset the kernel.  I missed the step of setting the api key as I was rapidly reading the text.   Since this is a personal machine, it would be ok to look for a file, read the api key from the file and if not there, then prompt for it and save it.  Would be nice if we could update a profile or have a key store or credentials store for other platforms.   it would just be a key name and key value that the user could add/change.  I'd protect it with their demo_user password and use that to encrypt/decrypt the values stored on disk.  I'd put that key store in the /Teradata directory.  .... BUT. for this demo, to get it out there, just do an obvious prompt for the API key. 

on:  Input In [12], in <cell line: 6>()
      3 query = """How many married customers have purchased the product?""" 

NameError: name 'source' is not defined

After the first time I got it, I restarted the kernel and cleared all outputs and clicked to run all steps and it failed again.  I double checked with another git pull and my machine was already up-to-date.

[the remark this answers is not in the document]
chetan-hirapara: > Need to say that the billing is on the left of the screen

This changes are in PR# https://github.com/Teradata/jupyter-demos/pull/339

Strange, both are separate PRs and branches but still #339 changes are showing in #340

chetan-hirapara: - Done

- User have to enter api key into prompt

chetan-hirapara: - Done, Please check it again.

DougEbel (commented):
The error handling helps but ... unfortunately, it is the main output I'm getting. You need to look at the generated SQL and Nudge OpenAI to generate valid SQL.  Example:  The following fails because they use COUNT as the column name:

SELECT profession, COUNT(*) AS count  FROM DEMO_MarketingCamp.Retail_Marketing  WHERE marital = 'Married' AND city = 'Phoenix'

Another failed with:  Syntax error: expected something between the 'DESC' keyword and the 'LIMIT' keyword.         ... I've never used Limit on teradata so not sure what is right. 

SELECT city, AVG(monthly_income_in_thousand) AS avg_income  FROM DEMO_MarketingCamp.Retail_Marketing  WHERE monthly_income_in_thousand IS NOT NULL GROUP BY city  ORDER BY avg_income DESC  LIMIT 1;

I tried to "hack" your def run_query by adding two lines (but didn't work)

 
    Do not use the column name count in the SQL query.
    
    Do not use limit key word in the SQL query.
    
To get the generated text, open up a SQL console and issue the following:
     Flush query logging with all;
     select sessionid, starttime, errorcode, errortext, querytext from dbc.qrylog where starttime > current_timestamp - interval '5' minute order by 1,2;

To get more of them, use the sessionid as the constraint instead of the starttime.

chetan-hirapara:
> The error handling helps but ... unfortunately, it is the main output I'm getting. You need to look at the generated SQL and Nudge OpenAI to generate valid SQL. Example: The following fails because they use COUNT as the column name:
> 
> SELECT profession, COUNT(*) AS count FROM DEMO_MarketingCamp.Retail_Marketing WHERE marital = 'Married' AND city = 'Phoenix'
> 
> Another failed with: Syntax error: expected something between the 'DESC' keyword and the 'LIMIT' keyword. ... I've never used Limit on teradata so not sure what is right.
> 
> SELECT city, AVG(monthly_income_in_thousand) AS avg_income FROM DEMO_MarketingCamp.Retail_Marketing WHERE monthly_income_in_thousand IS NOT NULL GROUP BY city ORDER BY avg_income DESC LIMIT 1;
> 
> I tried to "hack" your def run_query by adding two lines (but didn't work)
> 
> ```
> Do not use the column name count in the SQL query.
> 
> Do not use limit key word in the SQL query.
> ```
> 
> To get the generated text, open up a SQL console and issue the following: Flush query logging with all; select sessionid, starttime, errorcode, errortext, querytext from dbc.qrylog where starttime > current_timestamp - interval '5' minute order by 1,2;
> 
> To get more of them, use the sessionid as the constraint instead of the starttime.

* Added few shots prompting to tackle the keywords like LIMIT, count etc.

chetan-hirapara:
@DougEbel - Please take the latest version and try to execute the question suggested in last section

DougEbel (approved):
OUTSTANDING!!!

PR #342, opened 2023-07-12
Query service

nupur-lal:
initial notebook for query service

### DIFF
--- a/UseCases/Query_Service/Query_Service.ipynb
+++ b/UseCases/Query_Service/Query_Service.ipynb
@@ -1,0 +1,143 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "id": "0a47b7fb-f3b4-4237-bf1b-6368aa544334",
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdin",
+     "output_type": "stream",
+     "text": [
+      "\n",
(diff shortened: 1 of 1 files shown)


DougEbel (approved):
In intro:
- unfinished sentence: ... with a Teradata driver is not practical or feasible, some
- strange bullet: Execution of SQL Queries from the browser.Teradata web-based SQL editors

In Getting Data, you say:  We will use query this table from Query Service.

You defined QS_HOSTNAME and SYSTEM_NAME, and they seemed to be redundant with the 2nd canceling out the first.  You list out 2 ports but assume people know HTTP vs HTTPS and then  come to a conclusion "Thus for our demo ..."   You should indicate that HTTP is a port used for web communication and HTTPS is the more secure version that is encrypted and uses digital certificates.  In your "Thus" statement you should indicate that we use HTTPS for better security. 


Based on the possibly redundant options, in the following prompt, I entered "local"  for the host name which didn't work (obviously)

This demo has lots of prompts. Unlike other demos, using shift+enter isn't a good recommended approach.  In your intro, you should have a subhead: Suggested keyboard usage for this demo:  and then suggest the best way to go through it. 

In your intro, provide bullets of the different connection methods you will use in the demo and very briefly, when they would be used (e.g. for a couple of queries).  

I'll approve this but request another pull request to improve it.

PR #346, opened 2023-07-17
bug fix for notebook

nupur-lal:
fix for the bug/error received during the run for read only notebooks in remaining_useful_life usecase


DougEbel:
Thank-you

PR #347, opened 2023-07-17
Added new usecase

ForbiddenDevil:
@DougEbel Please review this usecase. I'd be glad to receive suggestions. Thank you.


DougEbel (commented):
I tried to make it work by adding a !pip install azureml but that was obviously not enough.  

I would suggest that all of the library stuff be at the top to make it easier for people to know what is required.

ForbiddenDevil:
@DougEbel I have made few changes and verified using a new machine. It should work now. Please check again.

DougEbel:
My testing of notebooks is:
reset kernel and clear all outputs
run all steps

1) you still had the pip installs commented out, so it failed first time
2) I said you need a very visible/obvvious prompt to get people to either enter their Azure credentials or run the info book on how to get credentials after the user resets the kernel with 0 0
3) when you get to 2. AzureML, maybe you can test to see if they responded to the prompts above to set the credentials, and put out a message (in a message box if possible or enclose message in ****** text box.  Tell them since they haven't added credentials, scan through the steps until you get to ______ 
4) you had a great prompt in teal:  Note: Please change the connection details in the _______ before moving forward. 
5) did you need to have a separate python program instead of something in a cell?   
6)  All import statements should be at the top of a notebook
7) Don't expect the user to uncomment "load the PMML file into Vantage".  if they don't maintain proper Python indentation, they will create errors.  Use a teal box to instruct them to only run the following if you haven't done the steps in Azure. 
8) you need a diagram at the top of the notebook that shows the steps that diverge when there is a need to run something in Azure.  You may need finer grained step numbers so you can make it clear that if you are going to get an AzureML account, you take the left branch and it begins with a link to your instructions notebook .  The right branch leaves at about the current step 2 and  ... oh ... your step numbers aren't consistent.   Not sure what step number the right branch starts at, but it has almost nothing until "#l oad the PMML file into Vantage" step

DougEbel:
First time executing failed splits the data in to train and test data ...  'test_df' is not a view. ... so I used data dictionary to delete all databases and objects from demo_user and start over. 

After you ask if user has azure subscription, you don't suggest that they can proceed but will be instructed when to skip steps.

Need bold (aqua) box telling them to skip step 5.    Later you have "Note: If you do not have AzureML or did not perform the above steps, ...."  You need to provide explicit guidance to those with and without Azure ML.   

I don't see any of the promised screen shots.   You promised:  "But what if I don't have AzureML? Don't worry, we will execute the steps before AzureML would be used, show you screen shots of what the AzureML user would be doing,"

In 7: Visualize the results, you start by getting info from TD_RegressionEvaluator.  The first column says the RMSE is 
11977.09201.   Hmmm is that good or bad. there is no context of that number.  If I go back to the 1st graph of Energy Units, that ranges from 10,000 to 25,000.  If an error of nearly 12,000 is what the RMSE is saying, that is pretty bad.   It is not enough to say what RMSE or R2 or ... is.  You need to state what it means. 

Your final graph is confusing.  It says the blue line is Actual Consumption, and the title says it is Energy Consumption prediction.  It is always good to show the actual and the prediction.  The dotted line of Linear Regression RMSE probably ok, but is labled as RMS.  ... or I don't follow what is intended. 

It looks like you are editing the training.py program from the notebook.  Would it be cleaner to write a file with the data and then read that from training.py.  That might mean you need to move both training.py and the data file to Azure. 

Even if we didn't have native AzureML calls in train.py, I would have merged the pull request except for the comments above.

DougEbel:
looks good

PR #351, opened 2023-07-25
Competitor proximity

nupur-lal:


DougEbel (approved):
This is a great demo.  I am merging it because it is functional and the main descriptions are fine.  There are areas in the demo where you go for 10-15 cells with little output or explanation.  Also take a final look at the graph output to ensure there is adequate description about what the picture means to educate the reader.

PR #364, opened 2023-08-07
flood_proximity_analysis initial notebook'

nupur-lal:


DougEbel:
This is a great demo and I'll merge it as is, but request some improvements in descriptions and question whether one of the maps is working. 

I think there needs to be a stronger story that this is for an insurance company versus a city planner for planning regulations.  You don't have all properties so as you look at details, why is there only one property in a several block area.   I don't think you have all street addresses from North Island, so later when you get the data, indicate it is for the covered properties. 

It took a bit of time to get the first map but was then amazed that its scope not only included new Zealand, but also on the other side of the globe in the area in Florida where I live showing streets only completed in the last few months. 

Not everyone viewing the demo may know what a Centroid is.  Simply refer to the centroid as the geographical center of the shape, then say you will label each with the Flood ID (FID) ... suggest that people mouse over the centroids. 

You selected a sample of 10,000 addresses but then you claim an analysis of all 1.6 million records in seconds ... not in this demo. Instead you could indicate on a full sized platform you could calculate on millions of records in seconds.

In 3.5 Use Client-side tools to create risk bands,   I only see 3 green dots and perhaps 50 red dots and no yellow. 

in 3.6 Heatmap Visualization, you don't make it clear what the "heat" refers to.  There are 2 maps but I was spending time trying to figure out what the heat was on the first map and you actually had something on the second. Describe what the first map shows and what the second shows You should also suggest clicking on shapes to select and zoom in to see detail, click on a marker to see the address.   And why are these maps narrower than the others?

In 3.7 Visualization in Plotly, I couldn't see anything except the map. I couldn't see risk bands with different colors even after I played with toe opacity setting.

PR #377, opened 2023-08-24
Dim

nupur-lal:
pull request for DIM after incorporating review comments from James


DougEbel (approved):
I went through the notebook and added more explicit narration on what the user should be looking for.

PR #378, opened 2023-08-24
incorporated review comments for flood proximity

nupur-lal:
Most of the points/zone etc not showing on the map was due to the subset of data I created. Created new subset of 60k records in the flood zone 7. 
Other comments/explanations added in the notebook.


DougEbel (commented):
The first map in 2.3 says you are putting flood zones on a map but all I see is a map with some sort of outline in the ocean. Are those the flood zones you are talking about?

Maps in 2.6 and 3.2 look identical. 

All maps need text after (or before) that explain what the map means and encourage the user to click/zoom/etc to understand what you say it means. 

All maps should have similar sizes. when we get to 3.6 and 3.7, the maps are short.

DougEbel (approved):
I clarified that most of the demo starting with 3 was focused on zone 7.  I updated some explanations after maps. On the last map, I changed the .3 color from lime to green because lime dots don't show up on the map background.  

There are some performance issues, especially when finding the addresses within the centroid of zone 7. I don't have the syntax in front of me, but the idea is to pre-filter the addresses based on whether that address falls within the MBR (minimum bounding rectangle) of the desired zone.  Determining if points fall within a geospatial shape requires lots of calculations.  The MBR has the min/max of the x & y so that the basic filtering can be done with 2 BETWEEN constraints.  For those passing that test, the more sophisticated point inside shape can be determined.

PR #381, opened 2023-08-28
Recommendations During Product Search - OpenSource model- Enhancement

chetan-hirapara:
Hi @DougEbel @ForbiddenDevil 

Please review the PR

Here the task description: https://github.com/Teradata/jupyter-demos/issues/376

Enhancement points from Doug's email:

- Inform user that Section 1 would take time upto 3-4 min to download the model and setup env.
- Add desc to section 5.
 _From_email: • When loading the embeddings in section 5, you have a “please be patient” line with a time estimate, but you should say why: 6,264 rows with 1028 columns each or nearly 6.5 million values. (I assume you evaluated alternate storage formats and data loading techniques to reduce the 2-4 minute load time)._
- Give 3 recommendations to the user instead of 2
- Add proper description in section 7.2. Inform user that what is happedning behind the seen. Check if display print statement along with progress bar
 _From email: "keep the user entertained by showing the current activity in a box below the progress bar that you now have in a “print” statement"_
- In section 7.2: suggest user to enter multi-word product names to get more interesting results than single word.
- Add desc to section 5 ; How the embeddings is looks like
 _From email: You might show the “skippers” 10 lines of products containing some text like ‘%cashew%’_


DougEbel (approved):
Looks good ... merged

PR #390, opened 2023-09-05
Sales forecasting uaf

ForbiddenDevil:
Added new demo as is with standardization for ClearScape platform. @DougEbel Please review. Thanks.


DougEbel (approved):
Wow ... 88 steps.  I'll merge it but request that you go back thru and make sure you have a statement after any graph that interprets something that should be of interest to the reader.  For example, in the Stationarity panel, you have 4 graphs to explain.  Don't expect the reader to know (e.g.) what a "Dataset Power Spectrum Results" means.  

For the tables, look at the numbers and explain if necessary.  e.g. You make a statement "If the mean and variance is somewhat similar then we consider that seasonality has been removed from the series" but when I look at the columns it looks like the MEAN_Weekly_Sales are an order of magnitude greater than the SD_Weekly_Sales. 

Don't just throw out number/graphs and move on.  ... educate your audience.

Issue #366, opened 2023-08-07
Recommendations During Product Search - OpenAI embeddings - Enhancement


chetan-hirapara:
**At this point in the demo, we need to process the data to get embeddings.**

- Users with an API key can enter their key and click Create embeddings.
- Users without an API key can load the pre-built embeddings.

**Steps for users with an API key:**

- Enter your API key.
- Click Create embeddings.
- Click Done to start using the embeddings.

**Steps for users without an API key:**

- Click Load data.
- Click Done to start using the embeddings.

**Using the embeddings for both types of users:**

- **Dropdown menu:** Users can select a product from a dropdown menu to get recommended products.
- **Text input:** Users can type in the name of a product to get recommended products.

DougEbel:
Looks like good proposed changes

PR #394, opened 2023-09-07
Enhancement changes for recommendations using OpenAI

chetan-hirapara:
Hi @DougEbel @ForbiddenDevil 

Please review this PR

### Here a task: https://github.com/Teradata/jupyter-demos/issues/366

**Description:**

**At this point in the demo, we need to process the data to get embeddings.**

- Users with an API key can enter their key and click Create embeddings.
- Users without an API key can load the pre-built embeddings.

**Steps for users with an API key:**

- Enter your API key.
- Click Create embeddings.
- Click Done to start using the embeddings.

 **Steps for users without an API key:**
- Click Load data.
- Click Done to start using the embeddings.

**Using the embeddings for both types of users:**

- Dropdown menu: Users can select a product from a dropdown menu to get recommended products.
- Text input: Users can type in the name of a product to get recommended products.


DougEbel (approved):
There are some great technical innovations in the usage of python in jupyter to build this demo. This is also a good educational piece describing the science.  I am troubled about the complexity of dealing with outside platforms like AzureML and OpenAI and wonder if we need a rating like they have for movies ...e.g. "MA for mature audiences".   There could be 2 dimensions to the rating: Technical depth and duration ... to prevent people from falling into the 80 step  Sales Forecasting demo. 

I'm also troubled about our use of AI because the skeptic might criticize that they could have found all of the products containing "Pecan" with a simple string search.  Hopefully the excellent educational material will distract them.

PR #424, opened 2023-11-07
CreditCard datapipeline notebook in python version

nupur-lal:
Teradataml python version notebook of existing notebook


nupur-lal:
compatible with teradataml 17.20.0.4 version (pass4)

PR #426, opened 2023-11-08
Mortgage calculator - financial bot

chetan-hirapara:
Hi @DougEbel 

Please review this PR


chetan-hirapara:
> Merged but **this demo is not stable**. Asking repeated questions can cause it to hang and there is no recovery except to restart the entire demo.

- Yes @DougEbel. We'll handle it in next phase of implementation.

PR #438, opened 2023-11-21
Updated header image - architecture

chetan-hirapara:
Updated header image - architecture


DougEbel:
there is an incorrect word at the bottom of the diagram. It says "closet products" and I think you mean "closest products"

PR #450, opened 2023-12-08
Parkinsons usecase changed to use sql intead of teradataml for Decisi…

shilpa-nalkande:
Parkinsons usecase changed for sql version of DecisionForest


DougEbel:
Tried running with teradataml 17,20.0.6 and got diagnostic:
TeradataMlException: [Teradata][teradataml](TDML_2008) Invalid type(s) passed to argument 'database_version', should be: str.

Also, (with my apologies for the tardy review), this version doesn't have the new branding.

PR #453, opened 2023-12-12
Updated review comments for ShipTimePred

shilpa-nalkande:
This is the new branch for Shipment Prediction. Needs review and merge if ok


DougEbel:
Problems with the graphs not being close in their prediction since they are using record number for the X axis and may be including both prep time and shipping time in the total time whereas the prep time may be dependent on the product category and the ship time may be dependent on the shipping method.

PR #457, opened 2023-12-14
Supply chain

nupur-lal:
@DallasBowden @DougEbel 
*changed ddl in cloud, corrected data for status begin loading in warehouse_oms
*added 2 plots showing default warehouse of customer and new assigned warehouse 
* added status timeline of order for all 3source system
* merged updated from Doug in previous commit


nupur-lal:
@DougEbel @DallasBowden rechecked everything and notebook is running fine hopefully we can merge this now.

PR #481, opened 2024-01-23
Pass6 changes for EnergyConsumptionForecasting

ForbiddenDevil:
@DallasBowden @pranaydave @DougEbel , please review.


DallasBowden (changes requested):
This statement is being presented as a fact that someone else said.  There's no reference or footnote to back up the claim.  I suggest we just delete it.  "It is said that increasing the accuracy of predictions by 0.1% can improve profitability".  
Everything else is fine for now.  I'm passing it along to Kevin for his review.

DallasBowden (changes requested):
Kevin said you did a fantastic job and the included an !
His comment:
"My only real comment is I'd like to see them introduce the td plot method they're using - explain the fact that visualization is a HARD problem at scale; and using this method allows us to visualize this data without data movement - punch up the great value of the feature."

DallasBowden (approved):
1. Re-Styled a paragraph.
Approved and Merged.

PR #485, opened 2024-01-24
Demoindex 0124 created.

DallasBowden:
Created new jupyter-demos.json file.  I also updated the Release Notes with a new Graphic in the Header.
I missed adding a couple of last minute things I did, like updating the graphic and renaming the #2 Models Ops demo and adding the 11 ModelOps notebook.


DallasBowden:
Commit not executed properly.

PR #494, opened 2024-01-30
Updated SalesForecastingARIMA for Pass6 changes

shilpa-nalkande:
Pass 6 changes done for Sales Forecast ARIMA


DallasBowden (commented):
I've sent this one to Kevin to review.

DallasBowden (changes requested):
From Kevin:
Overall - fantastic improvements, but I feel like we're really doing ourselves a disservice by only showing it forecasting a single series, when the main selling point is we can do multiple series in one go.
under "Why Vantage" - I think it implies forecast hundreds/thousands of series at once (the main benefit) - missing the word "series" - "forecast hundreds/thousands at once".  I would also recommend putting that value first before explaining all it does, as that is the main benefit of UAF in that it scales.
under "Basic analysis of the dataset" I would recommend some explanation that we're doing this in-db, including the plotting so it can scale.
Nice explanation on the join section!!!
Under Section 6 - still having a hard time with the ARIMA explanation ("ARIMA is a notation for (p,d,q) where the parameters are filled with values to indicate the ARIMA model being used for the purpose of training, testing and forecasting.")

DallasBowden (changes requested):
From Kevin:
Much improved - great value props.

Very nice explanation about visualizations and how we can do the heavy processing in-database!!!
Although the same text is copy/pasted in the Correlation matrix section - maybe instead of a complete restate, say "as above, we are performing these calculations in-database and only retrieving the summary data for the visualization" or something like that
Nice explanation on VAL
Model Training - copy/paste of the point in #1 again.
Under Model Validation - this doesn't make sense here, since the data has already been transformed this way - "Similar to GLM, we standardize input features, such as using ScaleFit and ScaleTransform, before using them in the function. The function takes only numeric features. We must convert the categorical features to numeric values before prediction." - could maybe use this in talking about operationalizing the model, where new data needs this, but it is a bit vague here.
Nicely done in talking about classification evaluator

shilpa-nalkande:
> From Kevin: Much improved - great value props.
> 
> Very nice explanation about visualizations and how we can do the heavy processing in-database!!! Although the same text is copy/pasted in the Correlation matrix section - maybe instead of a complete restate, say "as above, we are performing these calculations in-database and only retrieving the summary data for the visualization" or something like that Nice explanation on VAL Model Training - copy/paste of the point in #1 again. Under Model Validation - this doesn't make sense here, since the data has already been transformed this way - "Similar to GLM, we standardize input features, such as using ScaleFit and ScaleTransform, before using them in the function. The function takes only numeric features. We must convert the categorical features to numeric values before prediction." - could maybe use this in talking about operationalizing the model, where new data needs this, but it is a bit vague here. Nicely done in talking about classification evaluator

**"These comments seem to be for some other use case. I have not used VAL, CorrelationMatrix etc in this use case . Can you confirm from Kevin?"**

DallasBowden (approved):
1. Capitalizations and Spelling.
2. I don't know what this statement means. I removed it from the Introduction, " Here we are owing to Analyze the dataset given.".
3. I'm leaving the last section as "Required Materials".  That title doesn't really fit.  In the Dataset section, it mentions a file without any indication of where to find the file.

Approved and Merged.

PR #499, opened 2024-02-01
Updated BrokenDigitalJourney with Pass6 changes

shilpa-nalkande:
Pass 6 changes for Broken Digital Journey

### DIFF
--- a/UseCases/Broken_Digital_Journey/Broken_Digital_Journey_Python.ipynb
+++ b/UseCases/Broken_Digital_Journey/Broken_Digital_Journey_Python.ipynb
@@ -24,7 +24,23 @@
    "source": [
     "\n",
     "\n",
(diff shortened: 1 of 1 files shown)


DallasBowden (changes requested):
I'm not getting a chart when I execute this cell:
image

shilpa-nalkande:
The cell is just the definition of the function and will be executed only when its called in the below  cells where we are trying to showcase the paths. I have also mentioned  "Here we are just defining the function which will be used below in the notebook to show different paths." We can make it bold and add a ** to identify also may be mention the sections in which it is getting used 
please let me know your thoughts.

DallasBowden (approved):
1. "$10s of millions in revenue" changed to "10s of millions of dollars".
2. A statement like this from American Express needs a reference.
3. Do we need to use a trademark symbol everytime we mention nPath?
4. Changed "Vantage's ClearScape" to "ClearScape".
5. Changed "Teradata VantageCloud" to "Teradata Vantage" in 4.1.
6. Conclusion: Changed to "Thus, the analytic capabilities of ClearScape Analytics".
7. Changed "Required Materials" to "Resources".

Approved and Merged.

PR #500, opened 2024-02-01
pass6 telco_customer_churn

nupur-lal:
pass 6 changes for telco_customer_churn


DallasBowden (changes requested):
This one result seems to be backwards.  I would suspect that DSL would be the reason that people churn, not having fiber.  Please double-check the result.
image

nupur-lal:
@DallasBowden chnaged the explaination of the plot. The data in the table is as per plot more number of customers in FiberOptic but they also have higher churn as compared to DSL customers.

DallasBowden (approved):
Change "Teradata's ClearScape Analytics" to "ClearScape Analytics".
Approved and Merged.

PR #505, opened 2024-02-05
pass6 traindelay changes

nupur-lal:


DallasBowden (changes requested):
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/db255058/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/db255058/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">

I made all these changes except for #7.  

1.   "Luckily" is not the proper word to use in the intro.  The paragraphy doesn't need this adverb or   a conjunction like "However". <br>    2. Edited last line of Introduction.<br>     3. We need a clearer explanation of what a "Virtual Dataframe" is   and emphasize that it leaves the data in Vantage. If we're always going to   mention the "shape" of the data then a short explanation of that   would also be appropriate.<br>     4. "Teradata DataFrame" should be "Vantage   DataFrame".  We need to be   consistent with capitalization.    "pandas dataframe" should be "Pandas DataFrame". <br>    5. Don't use "Vantage's ClearScape Analytics".  "ClearScape Analytics" can be   used on it's own. <br>    6. I don't see any indication in the Introduction explaining what this demo   will accomplish.  <br>In section 3, the   word travel is used -- this would be more commonly referred to as a   trip.     <br>7. We need to always use a higher resolution screen capture.  The graphic in section 3. is too pixelated   and blurry.<br>     8. Section 4.  "Data Preparation   using the Massive Parallel Processing of Teradata".  This is the old way of referring to the   database.  We have to remember that   Teradata is used as a company name and not a product.  I still catch myself doing it all the time.  The graphic needs to be rebuilt with   Teradata Vantage or just Vantage and the JupyterLab and python graphic needs   to be spaced properly. <br>    9. There was an extended character at the bottom of the notebook in the   Reference Links.
--




</body>

</html>

DallasBowden (approved):
Removed the work "Luckily" from the intro.  Added 2024 to the Copyright.

PR #512, opened 2024-02-08
Updated Pass6 changes for Customer Behavior Analysis

shilpa-nalkande:
Created SQL notebook instead of PY_SQL  and included image of nPaths from the python version


DallasBowden (approved):
Approved and Merged.

PR #516, opened 2024-02-09
Update Generative_Question_Answering_Python.ipynb

JanethL:
Must define the Open AI model otherwise it will default to using 'text-davinci-003' which has been deprecated resulting in an error.  I recommend using model="gpt-3.5-turbo-instruct" as it has had the best results when I tested.


adamtwo (approved):

adamtwo:
@DougEbel , @DallasBowden  this change doesn't need any adjustments in CSAE. Since it's a break-fix, I recommend you merge ASAP.

DallasBowden:
I've merged this request, but I would like to know how to switch and view the changes using these names with colons in them:
Teradata:main
JanethL:patch-1

PR #568, opened 2024-03-29
Read only version for 4D Analytics Usecase

shilpa-nalkande:


DallasBowden (approved):
The notebooks works now.

DallasBowden (approved):
Both open from the navigator in JupyterHub, but not from the menu item.  I'm assuming I need to merge the pull request before they're commited so the read-only link works from demo.index.

PR #570, opened 2024-03-31
update gitignore to ignore demo_log and userinfo.db

DougEbel:


DougEbel:
leaving branch for review, but has been merged and can be deleted

PR #626, opened 2024-05-17
New notebook showing the teradatamlspk usage 

shilpa-nalkande:
Below GRANT needs to be included in the machine image before merging the notebook
GRANT CREATE EXTERNAL PROCEDURE on SYSUIF to demo_user ;


DallasBowden (changes requested):
1. I don't see a Use Case embedded in this notebook.  
2. It's missing a "Why Teradata" section at the top.
3. Section 6.  These values are hard to interpret. 
image  Since this notebook is about Predicting House Prices, can we have a chart that highlights those predictions?

DallasBowden (approved):
I have moved this folder from the UseCases Directory into the Getting_Started Directory.

PR #637, opened 2024-05-30
Updated the relative paths and the install command for teradataml

shilpa-nalkande:
Updated relative path and install and uninstall teradataml latest version


DallasBowden (approved):

PR #638, opened 2024-05-30
Updated the column names

shilpa-nalkande:
column name error


DallasBowden (approved):
The notebook executes as expected.

PR #644, opened 2024-06-13
Created another notebook with osml and trustedai code

shilpa-nalkande:
Created notebook with osml and lime for showing trusted ai


DallasBowden (approved):
The cell with the 3 !pip installs was chanded to 2 cells.  If teradataml does not need to be upgraded to 20.0.0.0, the lime and scikit-learn packages still need to be installed.  The downgrading of teradataml and scikit-learn was upgrading the numpy version.  The statement to downgrade it has been added to the cleanup process.  Added Beta Paragraph.

PR #648, opened 2024-06-19
bug fix for TD_ColumnTransformer

nupur-lal:
TD_ColumnTransformer sql was throwing error "Unable to run SQL: Unable to run SQL query: Database reported error:9134:CAF_MEM_API: Memory current in use plus requested allocation is less than 0." in db versions 23 , 25 and 26.
Re wrote the sql to clear the error.


DallasBowden (approved):

PR #657, opened 2024-07-02
Tdml20

nupur-lal:


DallasBowden (approved):
Matching new production image with latest TD and teradataml.

PR #675, opened 2024-07-26
NIGHTLY Playwright job

sreeramnitin:
And also input for manually running tests by providing coma separated file paths.


.github/workflows/ci.yml
```diff
@@ -9,26 +12,31 @@ on:
         options:
           - CI
           - PROD
```
adamtwo, line 14: How is this going to work when triggered from cron? Is it going to default to CI?

adamtwo (changes requested):

adamtwo (approved):

PR #684, opened 2024-08-02
add requirements.txt, update architecture diagram, fix formatting and…

JanethL:
- Updated the font and removed the light blue color in intro
- added a section noting to restart the kernel
- Moved packages to requirements.txt
- Headers updated to reflect 20 px and subheaders 18 px font 16 px
- Updated architecture diagram
- added input for region name


DallasBowden (approved):
We added an additional library to the requirements.txt file.
We added try except statements around the database load and drop statements so the users will get more feedback than just ''.
Added the Copyright statement to the bottom.

PR #685, opened 2024-08-05
add static populated page

JanethL:
Adding populated static page for a Read Only Version


DallasBowden:
Our Pop ReadOnly files go into a different branch, "origin/pass5_readonly".  I've pushed it there so now I just need to update the demo.index with a link to it.  

We don't need to commit this pull request so I'm deleting it.

PR #705, opened 2024-09-09
Fixed issue with suffix in Customer_reviews_analysis_using_GenAI

chetan-hirapara:
Fixed issue with suffix in Customer_reviews_analysis_using_GenAI


DallasBowden (approved):
Removed the Beta Paragraph.
Executes as Expected.

PR #708, opened 2024-09-10
Fixed customer id in mortgage calc

chetan-hirapara:
Fixed customer id in mortgage calc


DallasBowden (approved):
Executes as expected.

PR #718, opened 2024-09-26
Update bedrock langchain text-to-sql demo

JanethL:
- use create_context() function to establish connection to Teradata
- added section for exploring the data using Dataframe() constructor
- Added query band


DallasBowden (approved):

PR #729, opened 2024-10-22
updated sorted display for result

nupur-lal:


DallasBowden (approved):
Updated the wording in Section 2.

PR #735, opened 2024-11-05
Osml salespred

shilpa-nalkande:
New notebook for Sales Prediction for advertising data


DallasBowden (approved):
Updated queryband, removed library installs at the end of the notebook.

PR #740, opened 2024-11-14
Adding support for kernel restart in playwright scripts

sreeramnitin:
Feature/csae 438


adamtwo (approved):
fix spelling in restartKernal

PR #750, opened 2024-11-22
Edited the code to remove boto3 and bedrock imports; added pip and pi…

VB255047:
…llow upgrade for wordcloud


DallasBowden (approved):
I changed the data load statements to use foreign tables.  I added yaml files to provide values for our Playwright test.

PR #761, opened 2024-12-12
Exercise notebook

nupur-lal:


DallasBowden (changes requested):
This PR has a folder under our new Recipes folder, but the foldering isn’t correct.  There are other subjects that will be in the Recipes folder so you can’t have an immediate folder called Exercise_Notebook.  And a Getting started notebook is still a getting started notebook – it should go with the other Getting Started notebooks.  If we want to do something in this Exercise notebook, let's be more specific about what it is and put it under the Recipes folder.

PR #764, opened 2024-12-17
Added Model artifacts which were removed

shilpa-nalkande:
added model artifacts


DallasBowden (approved):
Nothing to execute.  Approved.

PR #818, opened 2025-01-23
New recipe notebook for Npath visualization widget

nupur-lal:


DallasBowden (approved):
Replaced the inclusion of the demo.gif file with a mp4 version.  This needs to be re-visited.

PR #827, opened 2025-01-28
new recipe notebook for nonlinearcombine fit and transform functions

nupur-lal:


DallasBowden (approved):
Capitalized F in function and added an s for it's plural form.  Updated HTML for dark mode. Approved.

PR #839, opened 2025-01-31
add new jupyter notebook for text to sql ai analyst agent with gemini

JanethL:
Requested for an innovation day with customer


DallasBowden (approved):
Question 10 doesn't return an answer.  I'll leave it in because it's a good example of what processing it will undertake to keep re-writing SQL to try to find an answer.

PR #840, opened 2025-02-02
Feature/hosted playwrigt reports

sreeramnitin:


adamtwo (approved):
It would probably make sense to demonstrate how this works to the team.

PR #861, opened 2025-02-12
new notebook for pack and unpack functions

nupur-lal:


DallasBowden (approved):
Approved. Removed the reference to AntiSelect in the introduction.  Cleared an output.

PR #864, opened 2025-02-13
Recipe for ACF and PACF

shilpa-nalkande:
The ACF() function calculates the autocorrelation or autocovariance of a time series.  The PACF() function is used to measure the degree of correlation between series sample points.


DallasBowden (approved):
Approved.  Added s to Function in the Title.

PR #865, opened 2025-02-13
new recipe notebook for stringsimilarity function

nupur-lal:


nupur-lal:
Calculates the similarity between two strings, using the specified comparison method (jaro, cosine etc)

DallasBowden (approved):
Approved.  Capitalized the F in Function in the title.

PR #867, opened 2025-02-13
new recipe notebook for random projection min components, fit and tra…

nupur-lal:
…nsform functions


nupur-lal:
Fit and Transform functions to reduce dimensionality using random matrices while preserving distances approximately. RandomProjectionMinComponents() function calculates the minimum number of components required for applying RandomProjection on the given dataset for the specified epsilon(distortion) parameter value.

DallasBowden (approved):
Approved. Capitalize F in Functions and made it plural.

PR #868, opened 2025-02-13
Update README.md

fs250044:
Fixed heading


fs250044:
@adamtwo or @DallasBowden , can one of y'all approve this in Doug's place? Thanks!

PR #876, opened 2025-02-19
Recipe for WhitesGeneral

shilpa-nalkande:
The WhitesGeneral() function checks for the presence of correlation among the residual terms after running a regression.


DallasBowden (approved):
Approved. Function was missing the F in the title.  Reworded a couple of paragraphs.

PR #885, opened 2025-02-24
New recipe for HoltWintersForecaster

shilpa-nalkande:
HoltWintersForecaster is a time series forecasting method that uses a combination of level, trend, and seasonality components to make predictions.


DallasBowden (approved):
Approved.

PR #886, opened 2025-02-24
New Recipe for MAMean

shilpa-nalkande:
The MAMean() function forecasts a user-defined number of periods into the future, that is the number of periods beyond the last observed sample point in the series.


DallasBowden (approved):
Approved.

DallasBowden (approved):
Approved.

PR #889, opened 2025-02-25
New Recipe notebook for DickeyFuller

shilpa-nalkande:
The DickeyFuller() function determines whether a set of time series data is stationary or nonstationary. It tests for the presence of a unit root in the data.


DallasBowden (approved):
Approved.

PR #890, opened 2025-02-25
new recipe notebook for ztest function

nupur-lal:


nupur-lal:
Tests the equality of two means under the assumption that the population variances are known.

DallasBowden (approved):
Approved.  Capitalized the f in functions and made it singular.

PR #891, opened 2025-02-26
New Recipe notebook for DurbinWatson

shilpa-nalkande:
The DurbinWatson() function determines if there is evidence of serial correlation between residuals in a residual series.


DallasBowden (approved):
Approved.

PR #892, opened 2025-02-26
New Recipe notebook for BreushGodfrey

shilpa-nalkande:
The BreuschGodfrey() function checks for the presence of serial correlation among the residual and error terms after running a regression associated with a fitted model.


DallasBowden (approved):
Approved.

PR #893, opened 2025-02-26
New Recipe notebook for BreushPaganGodfrey

shilpa-nalkande:
The BreuschPaganGodfrey() function is used to detect the presence of heteroscedasticity in regression analysis.  It checks for heteroscedasticity using one or more variables among the residual terms after running a regression.


DallasBowden (approved):
Approved.

PR #894, opened 2025-02-26
new recipe notebook for regression evaluator

nupur-lal:


nupur-lal:
Computes metrics to evaluate and compare  multiple regression models and summarizes how close predictions are to their expected values.

DallasBowden (approved):
Approved.  Capitalized the F in function.

PR #896, opened 2025-02-27
New Recipe notebook for CumulPeriodogram

shilpa-nalkande:
The CumulPeriodogram() function is used to perform a cumulative periodogram statistics test, also known as Bartlett's test, to determine the optimal data model. The function is usually executed against the residuals produced during the second 'model validation' exercise, meaning against the 'in-sample' forecasted data points.


DallasBowden (approved):
Approved.

PR #898, opened 2025-02-27
new recipe notebook for glm and tdglmpredict functions

nupur-lal:


nupur-lal:
Create a Generalized Linear Model on the dataset and predict the values using the predict function

PR #899, opened 2025-02-27
New Recipe notebook for SignifResidmean

shilpa-nalkande:
The SignifResidmean() function is a significance of residual mean statistics test to determine the optimum data model. The function uses the residuals generated during model validation that used the forecasted data points.


DallasBowden (approved):
Approved.

PR #904, opened 2025-02-28
New Recipe notebook for SignifPeriodicities

shilpa-nalkande:
The SignifPeriodicities() is a statistical test to determine if there exist any significant periodicities (seasonal cycles) in the residual series. This test can be used with any residual series, like the one produced by ArimaEstimate with fitpercentage(100) or ArimaValidate.


DallasBowden (approved):
Approved.

PR #906, opened 2025-02-28
new recipe notebook for glmpersegment and glmpredictpersegment functions

nupur-lal:


nupur-lal:
Create and predict GLM model for each segment using partition-by-key function to create a single model for each partition.

DallasBowden (approved):
Approved.  Capitalized the F in functions.

PR #907, opened 2025-02-28
New Recipe notebook for Portman

shilpa-nalkande:
The Portman() function (Portmanteau test) uses a series of tests to determine whether the residuals can be classified as zero mean, no evidence of serial correlation, or the residuals exhibit homoscedastic variance. These residuals are also referred to as white noise.


DallasBowden (approved):
Approved.

PR #908, opened 2025-03-03
New Recipe notebook for SimpleExp

shilpa-nalkande:
The SimpleExp() function uses simple exponential smoothing for the forecast model for univariate data. It does not use seasonality or trends for the model.


DallasBowden (approved):
Approved.  Capitalized the F in function.

PR #911, opened 2025-03-03
New Recipe notebook for MInfo

shilpa-nalkande:
MInfo() displays a result set containing metadata about each matrix. Each input payload produces the output variables for MIN_MAG, MAX_MAG and AVG_MAG. The output varies depending on the number of input payloads.


DallasBowden (approved):
Approved.  Capitalized the f in function.

PR #914, opened 2025-03-03
new recipe notebook for vectordistance function

nupur-lal:


nupur-lal:
Computes similarity or dissimilarity between two vectors in multi-dimensional space. It also supports distance/similarity computation for embeddings. The distance between vectors is usually calculated using a distance metric, such as Euclidean, Manhattan, DotProduct, Minkowski, or Cosine.

DallasBowden (approved):
Approved. Capitalized the f in function.

PR #916, opened 2025-03-03
new recipe notebook for ROC function

nupur-lal:


nupur-lal:
A receiver operating characteristic (ROC) curve shows the performance of a binary classification model as its discrimination threshold varies.

DallasBowden (approved):
Approved. Capitalized the f in function.

PR #917, opened 2025-03-03
new recipe notebook for classification evaluator function

nupur-lal:


nupur-lal:
Evaluate and produces various metrics of classification model based on its predictions on the data.

DallasBowden (approved):
Approved. Capitalized the f in funtion.

PR #926, opened 2025-03-11
fix hyperparameter parsing

GabrielCU:


DallasBowden:
This PR was not created in the foldering/repository the way I'm accustomed to working with them.  I've made the updates in another notebook and pushed it into main.

PR #945, opened 2025-04-08
Teradata end-to-end usecase for Telco

shilpa-nalkande:
Please let me know in case we should change the folder name or the Usecase name


DallasBowden (approved):
Approved.  Bypassing Playwright issues.

Issue #950, opened 2025-04-24
Fix json issue in diabetes H2o version demo


chetan-hirapara:
Fix json issue in diabetes H2o version demo

PR #951, opened 2025-04-24
Fixed json issue in diabetes H2o version

chetan-hirapara:
Fixed json issue in diabetes H2o version

issue: https://github.com/Teradata/jupyter-demos/issues/950


DallasBowden (approved):
Approved.  Updated the date in the Copyright notice.

PR #956, opened 2025-05-06
Basketball

nupur-lal:
new notebook for NBA data analysis using RAG & LLM


DallasBowden (approved):
Updated agent_executor to only return the answer: verbose=False
Added .yaml file.

PR #959, opened 2025-05-15
Fixed the issue with null json in Diabetes with H2o

chetan-hirapara:
Fixed the issue with null json in Diabetes with H2o

Task# https://github.com/Teradata/jupyter-demos/issues/950


DallasBowden (approved):
Approved.

PR #968, opened 2025-06-16
Added Dark-Theme support for the Recipe, Resume Analyzer, and Knee Replacement Use Case

patalvarado:
Modified the .ipynb files under Recipes, Resume Analyzer, and Knee Replacement use case directories to remove hard-coded colors to work with JupyterLab Dark Theme as well as JupyterLab Light Theme.


nupur-lal:
created new pr#1117 instead as some of the notebooks are already updated for dark theme

PR #981, opened 2025-07-31
Added Smartwatch - Healthcare Use Case

JasonD13:


DallasBowden (changes requested):
I like the proposed use case for this demo, but the supporting execution and data along with loading the data isn't there.  It needs to be more than a link to a Superset dashboard.
Also, please don't include "color:#00233C" in any of the html.

It needs to show this: "Leverage data from widely available smartwatch sensors (e.g., ECG, BVP, Accelerometer) to detect stress using machine learning models. Extract block-wise statistical features using Vantage’s in-database processing capabilities to ensure scalable and parallelized analysis. Augment stress detection by integrating sleep stage classification, enabling holistic health insights. The end-to-end solution allows for identifying high-stress intervals, correlating them with sleep and activity behavior, and ultimately enabling personalized wellness interventions."

PR #992, opened 2025-08-18
Servicenow path

nupur-lal:
new notebook on npaths with servicenow data


DallasBowden (approved):
Updated section headings that jumped from 4. to 6.

PR #994, opened 2025-08-25
Chatbot tdvs - latest

Ashwin-ap:


DallasBowden (changes requested):
Can we include a set of PDF files and structure the notebook so they can be used by default or still provide the user with the ability to pick their own?  We should also provide a few guidelines about what should be in the pdf files.

DallasBowden (approved):
I've added a few additional comments. I've updated the way we name the Vector Store so it's more directly associated with the user. I've added some logic for testing if objects already exist.

PR #995, opened 2025-08-25
getting started notebook for vector store

nupur-lal:


DallasBowden (approved):
Updated statements about assigning permissions for other users. Change the vector store names to include the database username first.

PR #1006, opened 2025-10-13
ComplaintsAnalysis VCL Demos - Refactor notebooks for dotenv auth and apply new formatting standards

Mohd-Hassan:
- Changed the authentication method to use **.env** files instead of **.json** for improved security and flexibility.

- Reformatted markdown cells within notebooks according to the latest convention to enhance readability and consistency.


DallasBowden (changes requested):
This is only for Complaint_Analysis_Customer360_VCL.ipynb

I've made the edits and have it working down to the cell where it's looking for the models on the local drive.  The first thing to check is that the pathing is correct.
Changed   Login Cell.     We're not including Kevin's OF Utils.    Updated Python version in library import cell.  We set the oaf environment name to the same as the database user   name.     Updated the OAF Environment Cleanup.     Removed 7.2 Stop the Cluster.     Removed Part 1 and Part 2 headings.     Added additional requirements file.

PR #1036, opened 2025-11-13
Teradataml widgets notebook update

saroop-samra:
These are latest notebooks along with requested changes in renaming (underscore instead of spaces), updating banner images, adding copyright at the bottom, and reorganizing the file structure so all data is in new data folder.


DallasBowden (approved):
I updated the pip instructions for these notebooks and updated a couple of pip installs. Also, I can't get the Widgets UI to work in the Analytic_Functions_Tutorial.ipynb notebook.

PR #1037, opened 2025-11-13
Refactor: direct API_Request call via SQL, boto3 instead of AWS CLI with updated formattings.

Mohd-Hassan:
- Updated Vantage workflow to invoke `API_Request()` directly through SQL .
- Replaced AWS CLI commands with `boto3` due to the ongoing issue with CLI commands.
- Refactored the notebook to align with the latest formatting standards.


DallasBowden (changes requested):
I've updated the markdown for dark mode and modified a couple of messages. I still can't get the TDApiClient.API_Request() to work.

shilpa-nalkande:
Tested with the new litellm changes. Works fine

PR #1042, opened 2025-11-24
Refactor: Replace AWS CLI usage with Boto3 for Bedrock GenAI demos

Mohd-Hassan:
Improves reliability of Bedrock interactions by using native Boto3 SDK instead of shell-based AWS CLI commands.


DallasBowden (changes requested):
Notebooks need to be edited since they no longer being used in a workshop.

Remove any references to getting the AWS keys from our Google Bucket.

Remove the configure_aws cell.

Ask for the access_key, secret_access_key and region the same way we do in other AWS notebooks.

Configure the boto3 call without the AWS access token.

Update the html styles for Dark Mode.

Update the queryband DEMO= values to be the same as the notebook filename.

DallasBowden (changes requested):
Please go back through these and remove all the references to color:#00233C. 
Please remove the button to get the aws keys.
Can we connect to our LiteLLM and pull the necessary aws credentials?

DallasBowden (changes requested):
Remove references to color:#00233C.
Remove the button to view the aws access credentials.
Can we connect his to our LiteLLM for the aws credentials?

PR #1083, opened 2026-02-26
Sentiment done, Speech recognition requires AWS Transcribe, so no use…

Ashwin-ap:
… of using litellm in that demo


DallasBowden (changes requested):
The cells in 3.6 will not generate the graphs.

PR #1104, opened 2026-05-06
Open api connectivity

nupur-lal:


DallasBowden (approved):
We should explain why the notebook needs to log in with the dbc user.

Can we come back around to this notebooks and get the openAI key from llmproxy?

PR #1109, opened 2026-06-11
Update LangChain Teradata version and notebook product  naming

JanethL:
Need to downgrade teradata-langchain version because VCL does not support Vector Store V2


DallasBowden (approved):
Updated pip installs. Updated business_name in the final validation sql. Added datainsights session aborts to cleanup.

